PR #4, opened 2020-09-02
Refreshing Nix Setup and Package Setup

CMCDragonkai:
@robert-cronin 

New nixpkgs revision means the shell.nix needs to change.


CMCDragonkai:
As a a demo library, the existing repo had way too much stuff that wasn't necessary as a template repository. Simplifying alot of things here.

CMCDragonkai:
When building `nix-build --attr application`. It's important to have a clean repo. The way node2nix does it is a bit strange atm.

Also while `nix-build --attr application` works. The `nix-build --attr docker` will produce an image that won't work. This is because the `/bin/typescript-demo-lib` is a symlink that doesn't get properly setup inside the container.

This is because node2nix generates a fairly weird artifact.

So it seems like node/js applications are quite strange.

At any case. This is working quite well otherwise.

CMCDragonkai:
Another thing is that `lint` is not meant to fix things. At least in CI/CD it's just meant to check, not actually fix things.

CMCDragonkai:
Fixing this now with:

```
npm run lint
npm run lintfix
```

The CI/CD should not be changing source code. It only runs checks.

Source code changes always requires a human review for now.

CMCDragonkai:
Webpack has been removed since as a library you don't want to bundle things.

CMCDragonkai:
Hey @nzhang-zh I've discovered a bug in our gitlab-runner container.

I don't think it has the `/usr/bin/env` file.

This causes things to fail, because alot of scripts assumes the idea that `#!/usr/bin/env node` can work.

This didn't happen on Python-Demo or Haskell-Demo, because those langauge environments rewrites the hashbangs.

Probably using `patchShebangs`.

But in this case, it doesn't happen.

CMCDragonkai:
Some of the changes here should be transported to js-polykey.

After some experimentation I'm beginning to think that js-polykey doesn't require a bundler at all since it is a CLI + library.

The Polykey main repo can use webpack.

But to retain the usage of `@js-polykey` or whatever then `module-alias` has to be used. But it has to be done in the `js` scripts.

Also notice how I'm compiling `src/bin` to `dist/bin`.

But I'm going to be reviewing https://github.com/MatrixAI/js-polykey/pull/84 next to see if we can move some ideas into there.

Furthermore docs generation isn't done yet here. But I reckon it's something that can be done separately.

CMCDragonkai:
I might explore using IFD (import from derivation) to separate the node2nix situation. Might make it easier to control. One thing is the ability to "generate a nix package" from a Node project which is what that's for. But it's important to ensure that source gitignore is working while... docker builds are also working. But that can be for later too.

CMCDragonkai:
Merging. @nzhang-zh fixed the above CI/CD issues.

But things that still require further exploration:

1. IFD (to be able to use node2nix properly in shell.nix (to get dev dependencies) and default.nix) - this can also mean we don't have to store the `/nix` files around.
2. Figure out why the docker tools build of the application failed, need to report this issue to node2nix upstream.

PR #11, opened 2020-09-10
Make main testable and see if we can remove dist from source

CMCDragonkai:
If docs and dist are both not in source and main is testable/importable, this makes the whole thing ALOT better!!!!


CMCDragonkai:
@robert-cronin For the next PRs into js-polykey, you should move to using similar `src/bin` setup and also `dist` isn't needed at all too. So that means removing webpack as well.

Issue #5, opened 2020-09-03
Explore using IFD when using node2nix


CMCDragonkai:
It could mean that `/nix` doesn't need to be stored. And allow the `default.nix` to use the `.package` while `shell.nix` to use `.shell`.

On the other hand, the `shell.nix` is still about using `npm` at the moment. So I wonder if that's better for development instead of trying to bring in the dev dependencies as nix dependencies as well.

Either way, the `default.nix` could make use of IFD.

CMCDragonkai:
In order to use Nix for everything, it would mean that `node2nix` IFD will need to generate all the dev dependencies when `shell.nix` imports it. And then enters that.

In that case we continue to use our own local cache.

CMCDragonkai:
Once we use IFD, we won't bother caching `node_modules` at all. Since all caching is done via nix and nix binary cache.

CMCDragonkai:
https://nixos.wiki/wiki/Import_From_Derivation

CMCDragonkai:
I think `shell.nix` shouldn't really be using `node2nix` as well. Really it just brings in node, and you continue to work with npm. This is because the workflow of JS stuff is just easier with `npm`. Doing this will also mean that one does not need to re-enter the shell to regenerate the `./nix` files. Instead that is just used when you need to invoke `nix-build` via IFD.

However `default.nix` can use IFD of the `node2nix`. This also means later one can use `nix-shell` of an IFD using `node2nix` inside the gitlab CI/CD so that all deps are using nix based dependencies when doing automated work. So that does mean CI/CD may actually be using `default.nix` and `pkgs.nix` directly instead of using `shell.nix` even when it uses `nix-shell`.

So after this, we eliminate the re-entering work and also remove the `./nix` files entirely.

CMCDragonkai:
Same thing can be done for Haskell-Demo.

@nzhang-zh can try this out tomorrow.

CMCDragonkai:
For pure libraries without executables, a shell.nix is sufficient as you don't do any distribution via Nix.

For libraries + plus executables, shell.nix should be independent of default.nix. Where default.nix is for packaging up as a Nix installable package, and of course being capable of of packaged up into a Docker container. However #6 is a problem.

So this demo lib will show  how to get a proper nix build soon. But for things like js-pagination is no default.nix is necessary.

CMCDragonkai:
@robert-cronin

CMCDragonkai:
[...]

PR #13, opened 2020-09-30
IFD Integration

CMCDragonkai:
Fixing #5


CMCDragonkai:
This almost works... 

I had to do a little trick.

So when node2nix runs, it runs `npm install`. 

When `npm install` runs, it actually runs: `preinstall`, `install` postinstall` and `prepare`.

If I add `npm run build` to `preinstall`. This ensures that the build runs before installation. This is required in order to build the `dist`, as I believe node2nix must inject a postinstall command that sets up the symlinks for the `bin`.

However this actually breaks `npm install`, since this then tries to compile before we have even installed any of the dependencies.

The proper place to put `npm run build` is in `prepare` script. This runs before prepublish, publish, postpublish.

But then `nix-build ./release.nix -A application` fails.

CMCDragonkai:
Relevant:

* https://stackoverflow.com/a/42076477/582917
* https://github.com/svanderburg/node2nix/issues/149

CMCDragonkai:
There are a couple solutions here.

1. Build `dist` ahead of time before running `nix-build`, then you don't use the `--development` flag, and you build only with the `dependencies` not `devDependencies`, this is how we used to do it, and we committed `dist` to source.
  * This defeats the purpose of IFD
  * Easy to do, just remove the `nix-gitignore.gitignoreSource`
2. Make nix do the actual build. This means `node2nix` to first get development dependencies... then run the build.
  * This is what we want, we want Nix to do the full building including the fetching of dependencies.
  * However the final package shouldn't include all of the dev dependencies... it should only have what is needed.

If all dependencies that are needed to build the dist are in `devDependencies`, then as soon as the build is done, another derivation can basically extract out the `dist` via a `publish` or `npm pack` and use that as the output. Then use `node2nix` for the normal dependencies.

Anyway this is not usable for now since the `npm install` is broken.

CMCDragonkai:
Need to hack node2nix so that the bin magic gets redesigned. Because however it is done, it should really be done AFTER `prepare` stage.

CMCDragonkai:
It must be reiterated that this is only relevant for Node packages that exposes executables. It is pretty useless for packages that don't expose executables.

For example if one were to be using webpack, node2nix is not relevant at all. Instead you might use Nix to run webpack instead and produce the dist directory directly. That will be interesting to try for the typescript-demo-spa.

CMCDragonkai:
So there's a way to create a minimal Nix package. That would be extracting the `dist` and `bin` and `node_modules` out of the `$out` built from the derivation.

Of course we would also need to inject in a `preinstall` command to build it. That way the derivation is actually building stuff into a temporary directory.

Anyway the build directory can then be dropped, since the files are copied, and there are no symlinks pointing back.

This package can then be packaged up for other situations too, and targeting different releases. It may even be possible to build our own docker representation too. And eventually other kinds of target platforms.

CMCDragonkai:
@robert-cronin this still requires some work to be done. Especially in order to properly integrate nix and especially if js-polykey can be packaged in docker. Leaving it for now as there are other higher priorities... but we will need to resolve this eventually.

sorki:
Neat! I've re-used your IFD and tried to align it a bit with `callCabal2nix` here https://gist.github.com/sorki/e23c3462f60e0b6a1eaa7d9fc3ac5480

I'm not going to use this for now due to `sandbox false` requirement, instead using a projects fork with nix files added https://github.com/otevrenamesta/matrix-appservice-slack/commit/9edd9153653434472300b2b0a0b2a4cd83172b01.

CMCDragonkai:
Hey if you see my previous comments you can see that node2nix doesn't work for docker. So that has to be fixed first.

nzhang-zh:
Regarding the docker build, I think the problems lies within the different handling of symlink'd directory between dockerTools and node2nix. 

In an application build, `$out/bin` is a symlink. 
```sh
~/m/T/result ❯❯❯ ll
total 9.0K
lrwxrwxrwx 25 root root 21 Jan  1  1970 bin -> lib/node_modules/.bin
dr-xr-xr-x  3 root root  3 Jan  1  1970 lib
```

In a docker build, `<container_root>/bin` is a plain directory.
```sh
[nix-shell:~/matrix.ai/TypeScript-Demo-Lib/container]$ ll
total 34
dr-xr-xr-x 2 root root 3 Jan  1  1970 bin
dr-xr-xr-x 3 root root 3 Jan  1  1970 lib
dr-xr-xr-x 3 root root 3 Jan  1  1970 nix
drwxr-xr-t 2 root root 2 Jan  1  1970 tmp
```

Therefore the symlink inside `bin/` breaks. 
```sh
# Application build
~/m/T/result ❯❯❯ ll bin/
total 512
lrwxrwxrwx 2 root root 64 Jan  1  1970 typescript-demo-lib -> ../@matrixai/typescript-demo-lib/dist/bin/typescript-demo-lib.js

# Docker build
[nix-shell:~/matrix.ai/TypeScript-Demo-Lib/container]$ ll bin/
total 1
lrwxrwxrwx 1 root root 64 Jan 15 11:56 typescript-demo-lib -> ../@matrixai/typescript-demo-lib/dist/bin/typescript-demo-lib.js
```

nzhang-zh:
And this can be fixed by turning on the `keepContentDirlinks` flag for `dockerTools.buildImage`.

```nix
    # When copying the contents into the image, preserve symlinks to
    # directories (see `rsync -K`).  Otherwise, transform those symlinks
    # into directories.
    keepContentsDirlinks ? false,
```

CMCDragonkai:
@nzhang-zh can you explain the difference between this and #16?

I'm guessing since we have 2 problems: docker builds and build staging of dev dependencies, we can have 2 separate MRs? One to simply solve the first, and the second requires more effort?

Both problems can use the IFD style.

nzhang-zh:
> the difference between this and #16?

#16's description has a summary of what it does. It is a copy of our previous workaround for `js-polykey`. It invokes `npm run build` at build stage with dev dependencies linked into the build directory as `node_modules`.

> This PR offers a simpler fix than IFD to the node2nix build problems using [this](https://github.com/svanderburg/node2nix#using-the-nodejs-environment-in-other-nix-derivations) as the main reference. 
> 
> Changes:
> - Add `-d` flag to `node2nix` call when generating nix files to include dev dependencies.
> - Symlinks node2nix's generated dependencies into build directory and run `npm run build`.
> - Creates a `bin/typescript-demo-lib` executable script that refers to the actual exe script in `dist` .
> - Copy `dist` and `bin` into `$out` in install phase.

>  we can have 2 separate MRs? One to simply solve the first, and the second requires more effort?

Sounds like a good plan. But the simple solution for docker actually works on top of the current IFD branch. 

> Both problems can use the IFD style.

The docker issue doesn't really require IFD. It is a `dockerTools` usage issue rather than `node2nix` issue. But the solution does build on top of the IFD branch.

CMCDragonkai:
Ok then @nzhang-zh get this PR ready to be merged first, squash and rebase and also clean up anything I didn't miss.

So this will solve the docker problem at least first, and will this solution here work for js-polykey as well?

I think @robert-cronin will need to first get rid of webpack and revert js-polykey back to a normal CLI-related project and synchronise its project metadata with typescript-demo-lib style.

Then we can test integration back into Matrix OS to ensure that js-polykey is still working properly and this time as a container...

CMCDragonkai:
Then after we merge this, we then redo #16 but based on IFD instead, and the goal there would be to produce the final executable output without the dev dependencies and only any non-dev dependencies.

CMCDragonkai:
Also can any #16 be synthesised into this PR?

nzhang-zh:
Another attempt. 
- Turned off `node2nix` generated `npm install` and added our own `npm run build` and `npm install` in `postInstall`
- Build is done using dev ifd and the dev dependency is replaced with production dependency after `npm run build` and `npm install` in `postInstall` as well

CMCDragonkai:
Squash rebase please.

CMCDragonkai:
Awesome!

@robert-cronin please review. We can start applying this to js-polykey.

PR #17, opened 2021-03-07
Optimising the default.nix to produce a minimal nix package

CMCDragonkai:
This updates the default.nix to produce a smaller package.

We realise that we only need `dist` after running the `npm run build`.

So we create another derivation that copies in only the relevant parts.

The result is a leaner output that has:

```
./result
├── bin -> /nix/store/i6lj4q204xzk19lff2w9f5adnvra0bwi-matrixai_typescript-demo-lib-1.0.0/lib/node_modules/.bin
└── lib
    └── node_modules
        └── @matrixai
            └── typescript-demo-lib
                ├── dist
                │   ├── bin
                │   │   ├── typescript-demo-lib.d.ts
                │   │   └── typescript-demo-lib.js
                │   ├── index.d.ts
                │   ├── index.js
                │   └── lib
                │       ├── Library.d.ts
                │       └── Library.js
                ├── node_modules
                │   └── uuid
                └── package.json
```

Removed all the `uuid` stuff just to make this message smaller.

Also note that the name is simplified to now `matrixai_typescript-demo-lib-1.0.0`. The `@` is removed and `/` is replaced with `_`. Apparently nix doesn't like having `@` in the nix store paths.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -20,13 +20,13 @@ let
       --composition $out/default.nix \
       --nodejs-${nodeVersion}
   '';
-  devPackage = (import (node2nixDrv true) { inherit pkgs nodejs; }).package;
-  prodDeps = (
+  # the shell attribute has the nodeDependencies, whereas the package does not
+  node2nixProd = (
     (import (node2nixDrv false) { inherit pkgs nodejs; }).shell.override {
       dontNpmInstall = true;
     }
   ).nodeDependencies;
-  drv = devPackage.overrideAttrs (attrs: {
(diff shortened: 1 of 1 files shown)


CMCDragonkai:
@DrFacepalm @scottmmorris @gideonairex

CMCDragonkai:
Tested it works:

```
nix-build ./release.nix -A docker
docker load --input ./result
docker run -it matrixai_typescript-demo-lib-1.0.0:f3wf1wxvkdbbmrhxx9rhr5qw2skbrah6
```

CMCDragonkai:
Technically the `package.json` isn't needed either, but we keep it there as useful metadata... it maybe useful for TypeScript-Demo-Electron.

PR #24, opened 2021-06-29
Adding Jest Test for stdout and console.log usage in CLI commands

emmacasolin:
Created a new function, demo-addition, that takes two numbers from the command line and adds them, then prints the result using console.log. This function utilises the NumPair class from a library called NumPair which stores two numbers. Tests were then created for this function using jest-test-process.

Once the function and tests were working for the console.log implementation the function and tests were then modified to use stdout instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "eslint-config-prettier": "^7.1.0",
     "eslint-plugin-prettier": "^3.3.1",
     "jest": "^26.6.3",
+    "jest-mock-process": "^1.4.0",
     "pkg": "^5.3.0",
     "prettier": "^2.2.1",
     "ts-jest": "^26.4.4",
--- a/tests/bin/typescript-demo-lib.test.ts
+++ b/tests/bin/typescript-demo-lib.test.ts
@@ -1,9 +1,64 @@
+import { mockProcessStdout } from 'jest-mock-process';
 import main from '@/bin/typescript-demo-lib';
+
+const uuidRegex = new RegExp(
+  '^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}',
+);
 
 describe('main', () => {
   test('main takes synthetic parameters', () => {
     // jest can also "spy on" the console object
     // and then you can test on stdout
     expect(main(['1'])).toEqual(0);
+  });
+  test('no input', () => {
+    const mockLog = mockProcessStdout();
+    main(['thing', 'thing']);
+    expect(mockLog).toHaveBeenCalledTimes(4);
+    expect(mockLog.mock.calls[0][0]).toBe('[]\n');
+    expect(mockLog.mock.calls[1][0]).toBe('new library\n');
+    expect(mockLog.mock.calls[2][0]).toMatch(uuidRegex);
+    expect(mockLog.mock.calls[3][0]).toBe('NaN + NaN = NaN\n');
+    mockLog.mockRestore();
+  });
+  test('adds 0 + 0', () => {
+    const mockLog = mockProcessStdout();
+    main(['thing', 'thing', '0', '0']);
+    expect(mockLog).toHaveBeenCalledTimes(4);
+    expect(mockLog.mock.calls[0][0]).toBe('[0,0]\n');
+    expect(mockLog.mock.calls[1][0]).toBe('new library\n');
+    expect(mockLog.mock.calls[2][0]).toMatch(uuidRegex);
+    expect(mockLog.mock.calls[3][0]).toBe('0 + 0 = 0\n');
+    mockLog.mockRestore();
+  });
+  test('adds 0 + 1', () => {
+    const mockLog = mockProcessStdout();
+    main(['thing', 'thing', '0', '1']);
+    expect(mockLog).toHaveBeenCalledTimes(4);
+    expect(mockLog.mock.calls[0][0]).toBe('[0,1]\n');
+    expect(mockLog.mock.calls[1][0]).toBe('new library\n');
+    expect(mockLog.mock.calls[2][0]).toMatch(uuidRegex);
+    expect(mockLog.mock.calls[3][0]).toBe('0 + 1 = 1\n');
+    mockLog.mockRestore();
+  });
+  test('adds 1 + 0', () => {
+    const mockLog = mockProcessStdout();
+    main(['thing', 'thing', '1', '0']);
+    expect(mockLog).toHaveBeenCalledTimes(4);
+    expect(mockLog.mock.calls[0][0]).toBe('[1,0]\n');
+    expect(mockLog.mock.calls[1][0]).toBe('new library\n');
+    expect(mockLog.mock.calls[2][0]).toMatch(uuidRegex);
+    expect(mockLog.mock.calls[3][0]).toBe('1 + 0 = 1\n');
+    mockLog.mockRestore();
+  });
+  test('adds 7657 + 238947', () => {
+    const mockLog = mockProcessStdout();
+    main(['thing', 'thing', '7657', '238947']);
+    expect(mockLog).toHaveBeenCalledTimes(4);
+    expect(mockLog.mock.calls[0][0]).toBe('[7657,238947]\n');
+    expect(mockLog.mock.calls[1][0]).toBe('new library\n');
+    expect(mockLog.mock.calls[2][0]).toMatch(uuidRegex);
+    expect(mockLog.mock.calls[3][0]).toBe('7657 + 238947 = 246604\n');
+    mockLog.mockRestore();
   });
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -3488,6 +3488,12 @@
         "@types/node": "*"
       }
     },
+    "jest-mock-process": {
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/jest-mock-process/-/jest-mock-process-1.4.0.tgz",
+      "integrity": "sha512-3LM1TyEaRKRjh/x9rZPmuy28r7q8cgNkHYcrPWtxXT3ZzPPS+bKNs2ysb8BJPVB41X5yM1sMtatvE5z2XJ0S/w==",
+      "dev": true
+    },
     "jest-pnp-resolver": {
       "version": "1.2.2",
       "resolved": "https://registry.npmjs.org/jest-pnp-resolver/-/jest-pnp-resolver-1.2.2.tgz",
--- a/src/bin/typescript-demo-lib.ts
+++ b/src/bin/typescript-demo-lib.ts
@@ -2,13 +2,28 @@
 
 import process from 'process';
 import Library from '../lib/Library';
+import NumPair from '../lib/NumPair';
 import { v4 as uuidv4 } from 'uuid';
 
 function main(argv = process.argv): number {
-  console.log(argv.slice(2));
+  // Print out command-line arguments
+  const argArray = argv.slice(2);
+  const args = argArray.toString();
+  process.stdout.write('[' + args + ']\n');
+
+  // Create a new Library with the value someParam = 'new library'
+  // And print it out
   const l = new Library('new library');
-  console.log(l.someParam);
-  console.log(uuidv4());
+  process.stdout.write(l.someParam + '\n');
+
+  // Generate and print a uuid (universally unique identifier)
+  process.stdout.write(uuidv4() + '\n');
+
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -19,7 +19,8 @@
     "test": "jest",
     "lint": "eslint '{src,tests}/**/*.{js,ts}'",
     "lintfix": "eslint '{src,tests}/**/*.{js,ts}' --fix",
-    "typescript-demo-lib": "ts-node -r tsconfig-paths/register src/bin/typescript-demo-lib.ts"
+    "typescript-demo-lib": "ts-node -r tsconfig-paths/register src/bin/typescript-demo-lib.ts",
+    "newFunction": "ts-node -r tsconfig-paths/register src/newFunction.ts"
```
CMCDragonkai, line 23: New CLI commands should go into the src/bin.

src/newFunction.ts
```diff
@@ -0,0 +1,6 @@
+function newFunction() {
+    let msg = "Hello World!";
+    console.log(msg);
+}
+
+newFunction();
```
CMCDragonkai, line 6: You should be exporting the function and importing in the CLI script.

CMCDragonkai:
Show me an execution of this on the terminal.

src/newFunction.ts
emmacasolin: Could you please clarify how this would be done? Where does the function get exported to and where does the CLI script get imported from?

CMCDragonkai:
Copy the existing src/bin/typescript-demo-lib.ts.

tests/bin/demo-addition.test.ts
```diff
@@ -0,0 +1,28 @@
+// this test fails -> attempting to check that the output of console.log is correct
```
CMCDragonkai, line 1: You have to use jest to spyon or mock the process.stdout.

Note that the command isn't using `console.log` to output data.

It's using `process.stdout` or `process.stderr`.

To simplify doing this, see: https://github.com/EpicEric/jest-mock-process

CMCDragonkai:
Try using `npm run lintfix` as well to reformat the code.

src/newFunction.ts
CMCDragonkai: Export it like:

```
export {
  newFunction
};
```

At the end of a typescript module.

Then import it in your command script like `import { newFunction } from '../lib/...';`

Try this so you can practice importing and exporting code.

If you look at js-polykey this is how we structure our code.

package.json
CMCDragonkai: Get the `bin` section filled out as well in `package.json`.

emmacasolin:
Still need to create working tests (see https://github.com/EpicEric/jest-mock-process)
The NewFunction files should have all been removed (changed to demo-addition in line with naming conventions as well as changed function to meet demo goals)

emmacasolin:
Implemented tests for demo-addition. I originally was mocking stdout to check that the output of the tests were correct, however I ended up deciding to mock console.log instead because the output there is simply the function's output without all the extra stuff that stdout adds in (e.g. console.log   []    at Object.main [as default] (src/bin/typescript-demo-lib.ts:8:11) in the case of the preexisting demo function) so I found mocking console.log a bit easier to track.

The code for all of this was borrowed from here: https://github.com/EpicEric/jest-mock-process
--> in general I understand the purpose of each line, bit I don't know exactly what every word does. I managed to make it work for my own purposes though so I think my understanding is good enough for now

tests/bin/demo-addition.test.ts
```diff
@@ -0,0 +1,55 @@
...
+>(
+  target: T,
+  property: M,
+  impl: I,
+): jest.SpyInstance<ReturnType<F>, ArgsType<F>> {
+  maybeMockRestore(target[property]);
+  return jest.spyOn(target, property as any).mockImplementation(impl);
+}
+
+const mockConsoleLog = () => spyOnImplementing(console, 'log', (() => ({})));
```
emmacasolin, line 27: I realised after pushing the tests to this branch that the code wasn't style compliant. I've since fixed almost everything with the latest commit, however I'm still having issues with this line, in particular `npm run lint` wants to change `() => ({})` to `) => ({}`, which breaks the code. I might try defining a function that's just an empty arrow function and putting that in there to see if that fixes the style issues.

tests/bin/demo-addition.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+type FunctionPropertyNames<T> = {
+  [K in keyof T]: T[K] extends (...args: any[]) => any ? K : never;
+}[keyof T];
+type ArgsType<T> = T extends (...args: infer A) => any ? A : never;
+
+function spyOnImplementing<
+  T extends typeof console,
+  M extends FunctionPropertyNames<T>,
+  F extends T[M],
```
CMCDragonkai, line 16: This is all unnecessary, you should not need to reimplement these functions at all.

Tests should be simple, all you do is import utilities from the outside and run a test.

These functions here are going to add maintenance overhead and won't be useful in the future.

tests/bin/demo-addition.test.ts
CMCDragonkai: This doesn't quite look right. Mocking the console.log is a simple thing, there's no need to create so many utility functions.

CMCDragonkai: Furthermore, when you test the main process, it's not even using console.log.

It's using STDOUT.

Therefore if you are using STDOUT, you must use the jest-mock-process, not the console.log mocking.

The process.stdout and console.log are 2 totally different things.

src/bin/demo-addition.ts
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+import process from 'process';
+import NumPair from '../lib/NumPair';
+
+function main(argv = process.argv): number {
+  const nums = new NumPair(parseInt(argv[2]), parseInt(argv[3]));
+  const sum = nums.num1 + nums.num2;
+  console.log(nums.num1 + ' + ' + nums.num2 + ' = ' + sum);
```
CMCDragonkai, line 9: Oh if you are doing this, then there's console.log being used.

But CLI commands should be using process.stdout not console.log.

tests/bin/demo-addition.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+  test('adds 0 + 0', () => {
+    const mock00 = mockConsoleLog();
+    main(['thing', 'thing', '0', '0']);
+    expect(mock00).toHaveBeenCalledWith('0 + 0 = 0');
+  });
+  test('adds 0 + 1', () => {
+    const mock01 = mockConsoleLog();
+    main(['thing', 'thing', '0', '1']);
+    expect(mock01).toHaveBeenCalledWith('0 + 1 = 1');
```
CMCDragonkai, line 43: Try this instead:

```
test('...', () => {
  const mockLog = mockConsoleLog();
  main([...]);
  expect(mockLog).toHaveBeenCalledWith('Browser log');
  mockLog.mockRestore();
});
```

The above is for when console.log is used.

Then change your script to using `process.stdout.write("...\n");` instead.

Then use the `mockProcessStdout();` instead.

CMCDragonkai:
@emmacasolin rewrite your PR to be like this:

Title: Adding Jest Test for stdout and console.log usage in CLI commands
Description: Write down a list of items/tasks that you did for this PR.

Make it more easily readable and understandable what the problem is.

CMCDragonkai:
This way we can actually incorporate your PR.

CMCDragonkai:
Also please rebase on master.

To do this, you must use:

```
git checkout master
git pull
git checkout fiddling-with-things
# at this point you'll hit conflicts
git rebase master
git push -f
```

Those conflicts will be problematic to fix.

Instead here's what you can do, get copy the `package.json` and `package-lock.json` from master here, so you have the same files as master.

Now instead of using the new command, use the existing `typescript-demo-lib.ts` command and use `process.stdout.write`. Look at the node docs for how to do this. Google "node process stdout".

Then test that using `jest-mock-process`.

You'll need to install the package with `npm install --save-dev jest-mock-process`.

emmacasolin:
There are now merge conflicts between this branch and master so I've copied the versions of package.json and package-lock.json from master and re-added the commands for demo-addition into package.json. I'm hoping this will resolve the conflicts

emmacasolin:
Re-adding the commands back in still leads to merge conflicts... so I'll need to find a way to make it work without them.

"Now instead of using the new command, use the existing typescript-demo-lib.ts command and use process.stdout.write. Look at the node docs for how to do this. Google "node process stdout"."

emmacasolin:
I now understand how to use jest-mock-process - installed using npm and so could remove helper functions from demo-addition.test.ts.

Was getting an error that mockConsoleLog was not callable, however I was missing the {} on the import line and that problem has now been fixed.

emmacasolin:
Was using `expect(mockLog).toHaveBeenCalledWith('Browser log');` to test output but this was incorrect - should be the actual expected output in place of `Browser log`.

emmacasolin:
This branch has now been rebased on master, I believe

src/bin/demo-addition.ts
CMCDragonkai: Try to add `process.stdout.write` as well. Then your tests should involve the `process.stdout.write` not `console.log`.

CMCDragonkai:
Please squash this branch down to 1 commit.

Use `git log` to  find the commit before the last commit you want to squash to.

Then copy the commit revision hash.

Use: `git rebase -i <COMMITHASH>`, and then change the `pick` to `squash` for all the commits EXCEPT the first commit.

Then save and exit your editor. (This might be vim or nano).

If this results in problems, you'll need to abort your rebase `git rebase --abort`.

Then instead use a soft reset. This means `git reset --soft <COMMITHASH>` I think this is how it works, or you might need to use the next commit hash instead.

This should undo all the last commits until that commit, and then allow you do a `git commit -m '<ENTER MESSAGE>'`.

Then if it looks good, that is you didn't reset the commit that already exists on master (check this by going to `git log`), then you can do a `git push -f`.

Remember this part is dangerous. And if you think you did it wrong, you can always reset back to this remote branch with `git reset --hard origin/fiddling-with-things`.

Unless you force pushed to this branch, in which case all the history would be lost on this branch. In that case you might just want to close this PR, and start a new PR and do it from a clean commit history branching off from master.

Make sure your repo is up to date with master.

CMCDragonkai:
BTW there's still a test failure.

CMCDragonkai:
The CI/CD shows this.

emmacasolin:
Tests were still failing on the CI/CD because I hadn't pushed the versions of package.json and package-lock.json that contained jest-mock-process to the branch yet - this has now been resolved

emmacasolin:
typescript-demo-lib.ts now runs the function that was contained in demo-addition, as well as using process.stdout.write for everything instead of console.log. I've pushed this to the branch so I now have no unstaged changes and can attempt to rebase again.

emmacasolin:
This branch has now been successfully rebased on master! With help from Joshua

emmacasolin:
Testing for typescript-demo-lib.ts will be more complex than the tests for demo-addition were as there are multiple calls to `process.stdout.write`. Instead of using `expect(mockLog).toHaveBeenCalledWith(output)` I will need to use `expect(mockLog.mock.calls[0][call number]).toBe(output for that call)`. Might also add a test for the number of times process.stdout.write was called.

emmacasolin:
Tests are done; ready to be merged

emmacasolin:
I think all of the writes to stdout can be tested in one go, however, this single test would need to use `.toMatch` rather than `toBe` since the third call is a randomly generated uuid (needs to be tested against a regex to ensure it matches the correct pattern, since the actual contents are random and can't be tested for directly). This means the single test won't be as reliable as the multiple tests, but it will be easier to maintain which makes up for that.

package.json
```diff
@@ -23,6 +23,7 @@
     "typescript-demo-lib": "ts-node -r tsconfig-paths/register src/bin/typescript-demo-lib.ts"
   },
   "dependencies": {
+    "jest-mock-process": "^1.4.0",
```
CMCDragonkai, line 26: This should go into `devDependencies`. Like `npm install --save-dev`. Can you move it in to the `devDependencies` section.

tests/bin/typescript-demo-lib.test.ts
```diff
@@ -1,9 +1,64 @@
 import main from '@/bin/typescript-demo-lib';
+import { mockProcessStdout } from 'jest-mock-process';
```
CMCDragonkai, line 2: Always put external imports above internal imports. So move this ahead of line 1.

PR #25, opened 2021-07-01
Integrate Vercel/Pkg for Release Distribution

CMCDragonkai:
This https://github.com/vercel/pkg seems to provide us a way of packaging CLI applications as an executable for distribution to Linux, Windows and Mac.

We need to try it here before PK. And furthermore, the `vercel/pkg` is not currently part of `nodePackages` attribute in Nix. So I imagine similar to Electron, there may be a need to have this done in Nixpkgs as there may be some things changed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,6 +4,7 @@ variables:
 stages:
   - check
   - build
+  - release
 
 image: registry.gitlab.com/matrixai/engineering/maintenance/gitlab-runner
 
@@ -33,6 +34,9 @@ nix-dry:
   script:
     - nix-build -v -v --dry-run ./release.nix --attr application
     - nix-build -v -v --dry-run ./release.nix --attr docker
+    - nix-build -v -v --dry-run ./release.nix --attr package.linux.x64.elf
+    - nix-build -v -v --dry-run ./release.nix --attr package.windows.x64.exe
+    - nix-build -v -v --dry-run ./release.nix --attr package.macos.x64.macho
 
 nix:
   stage: build
@@ -40,7 +44,31 @@ nix:
   script:
     - >
         nix-build ./release.nix
+        --max-jobs $(nproc)
         --attr application
         --attr docker
+  only:
+    - master
+
+packages:
+  stage: release
+  interruptible: true
+  script:
+    - >
+        nix-shell -I nixpkgs=./pkgs.nix --packages git gitAndTools.gh --run '
+        builds="$(nix-build \
+          --max-jobs $(nproc) --cores $(nproc) \
+          ./release.nix \
+          --attr package.linux.x64.elf \
+          --attr package.windows.x64.exe \
+          --attr package.macos.x64.macho)";
+        commit="$(git rev-parse --short HEAD)";
+        gh release \
+          create "$commit" $builds \
+          --title "Build-$(date -u +"%Y-%m-%dT%H:%M:%SZ")" \
+          --prerelease \
+          --notes "" \
+          --repo MatrixAI/TypeScript-Demo-Lib;
+        '
   only:
     - master
--- a/default.nix
+++ b/default.nix
@@ -1,57 +1,26 @@
 { runCommandNoCC
-, nix-gitignore
-, nodejs
-, nodePackages
-, pkgs
-, lib
+, callPackage
 }:
 
 let
-  src = nix-gitignore.gitignoreSource [".git"] ./.;
-  nodeVersion = builtins.elemAt (lib.versions.splitVersion nodejs.version) 0;
-  node2nixDrv = dev: runCommandNoCC "node2nix" {} ''
-    mkdir $out
-    ${nodePackages.node2nix}/bin/node2nix \
-      ${lib.optionalString dev "--development"} \
-      --input ${src}/package.json \
-      --lock ${src}/package-lock.json \
-      --node-env $out/node-env.nix \
-      --output $out/node-packages.nix \
-      --composition $out/default.nix \
-      --nodejs-${nodeVersion}
-  '';
-  # the shell attribute has the nodeDependencies, whereas the package does not
-  node2nixProd = (
-    (import (node2nixDrv false) { inherit pkgs nodejs; }).shell.override {
-      dontNpmInstall = true;
+  utils = callPackage ./utils.nix {};
+  drv = runCommandNoCC
+    "${utils.basename}-${utils.node2nixDev.version}"
+    {
+      version = utils.node2nixDev.version;
+      packageName = utils.node2nixDev.packageName;
     }
-  ).nodeDependencies;
-  node2nixDev = (import (node2nixDrv true) { inherit pkgs nodejs; }).package.overrideAttrs (attrs: {
-    src = src;
-    dontNpmInstall = true;
-    postInstall = ''
-      # The dependencies were prepared in the install phase
-      # See `node2nix` generated `node-env.nix` for details.
-      npm run build
-
-      # This call does not actually install packages. The dependencies
-      # are present in `node_modules` already. It creates symlinks in
-      # $out/lib/node_modules/.bin according to `bin` section in `package.json`.
-      npm install
-    '';
-  });
-  name = "${builtins.replaceStrings ["/" "@"] ["_" ""] node2nixDev.packageName}-${node2nixDev.version}";
-  drv = runCommandNoCC name {} ''
-    mkdir -p $out/lib/node_modules/${node2nixDev.packageName}
+    ''
+    mkdir -p $out/lib/node_modules/${utils.node2nixDev.packageName}
     # copy only the dist
-    cp -r ${node2nixDev}/lib/node_modules/${node2nixDev.packageName}/dist $out/lib/node_modules/${node2nixDev.packageName}/
+    cp -r ${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}/dist $out/lib/node_modules/${utils.node2nixDev.packageName}/
     # copy over the production dependencies
-    if [ -d "${node2nixProd}/lib/node_modules" ]; then
-      cp -r ${node2nixProd}/lib/node_modules $out/lib/node_modules/${node2nixDev.packageName}/
+    if [ -d "${utils.node2nixProd}/lib/node_modules" ]; then
+      cp -r ${utils.node2nixProd}/lib/node_modules $out/lib/node_modules/${utils.node2nixDev.packageName}/
     fi
     # create symlink to the deployed executable folder, if applicable
-    if [ -d "${node2nixDev}/lib/node_modules/.bin" ]; then
-      cp -r ${node2nixDev}/lib/node_modules/.bin $out/lib/node_modules/
+    if [ -d "${utils.node2nixDev}/lib/node_modules/.bin" ]; then
+      cp -r ${utils.node2nixDev}/lib/node_modules/.bin $out/lib/node_modules/
       ln -s $out/lib/node_modules/.bin $out/bin
     fi
   '';
--- a/package-lock.json
+++ b/package-lock.json
@@ -977,6 +977,15 @@
       "integrity": "sha512-OPdCF6GsMIP+Az+aWfAAOEt2/+iVDKE7oy6lJ098aoe59oAmK76qV6Gw60SbZ8jHuG2wH058GF4pLFbYamYrVA==",
       "dev": true
     },
+    "agent-base": {
+      "version": "6.0.2",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz",
+      "integrity": "sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==",
+      "dev": true,
+      "requires": {
+        "debug": "4"
+      }
+    },
     "ajv": {
       "version": "6.12.6",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
@@ -1035,6 +1044,22 @@
       "requires": {
         "normalize-path": "^3.0.0",
         "picomatch": "^2.0.4"
+      }
+    },
+    "aproba": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
+      "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
+      "dev": true
+    },
+    "are-we-there-yet": {
+      "version": "1.1.5",
+      "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
+      "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
+      "dev": true,
+      "requires": {
+        "delegates": "^1.0.0",
+        "readable-stream": "^2.0.6"
       }
     },
     "arg": {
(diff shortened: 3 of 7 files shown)


CMCDragonkai:
The vercel/pkg has 2 interesting dependencies:

```
    "pkg-fetch": "3.2.1",
    "prebuild-install": "6.0.1",
```

The `pkg-fetch` appears to have pre-built nodejs binaries used by pkg. This is similar to how our electron-forge in Nixpkgs (https://github.com/NixOS/nixpkgs/pull/115230) packages up electron applications by fetching pre-built electron releases and just bundling our application with it. Note that the lack of a proper electron ecosystem in Nixpkgs means that it's a bit adhoc.

Our TypeScript-Demo-Electron uses `utils.nix` to provide specific Electron binaries to be used instead of making electron-forge download them. This is mainly because during the build phase we cannot have other tools downloading things, these must all be specified as part of Nix. This is what makes Nix builds deterministic (mostly). To achieve this we had to use `ELECTRON_KIP_BINARY_DOWNLOAD = "1"` and `electron_zip_dir` which is specified in the `release.nix`.

We may need to do something similar here with vercel/pkg if we want vercel/pkg to be used within a release.nix build phase.

CMCDragonkai:
This PR is also updating the nixpkgs.nix to match the 21.05 release on our nixpkgs overlay. OS apps obviously doesn't use our private overlay for now.

CMCDragonkai:
Hit an error from running pkg https://github.com/vercel/pkg/issues/1242. A stream write error, which doesn't seem to be based on any operating system.

CMCDragonkai:
I tried to run on Windows, the EPIPE error isn't occurring. So maybe it does have something to do with NixOS or Linux. We can also try this in a VirtualBox.

CMCDragonkai:
Upon running the `pkg` command, it asks to install several targets. These targets are stored in the `PKG_CACHE_PATH`.

This defaults to:

```
»» ~
 ♖ tree .pkg-cache                                                                                                                               ⚡(master) pts/6 21:23:24
.pkg-cache
└── v3.1
    ├── fetched-v14.17.0-linux-x64
    ├── fetched-v14.17.0-macos-x64
    └── fetched-v14.17.0-win-x64

1 directory, 3 files
```

So this is something we can use as the equivalent of the `electronZipDir` when doing this during a nix build.

CMCDragonkai:
The built executables weigh in about 33 - 36 MiB for a simple hello world program. It's probably the entire node runtime plus the executable.

CMCDragonkai:
Ok tested on WSL as well, this builds a Linux executable that works on WSL (Ubuntu), and is also able to build the Windows executable which works via Powershell as well. The result is in the current directory:

```
typescript-demo-lib-linux
typescript-demo-lib-win.exe
```

CMCDragonkai:
The reason for choosing node14 is that node14 is the LTS release, and it is also the node that we use when we are in the nix-shell.

CMCDragonkai:
Running NixOS in virtualbox with an older kernel and older NixOS distribution version still fails with the same EPIPE error. I suspect this might mean that it may fail in NixOS generally or even any Linux distribution (barring the WSL Ubuntu distro).

CMCDragonkai:
I copied the linux executable that was built on WSL to NixOS to see if the ideas behind patchelf work here (https://nixos.wiki/wiki/Packaging/Binaries).

First thing is to see what interpreter path it's using (which is the dynamic link loader):

```
> patchelf --print-interpreter ./typescript-demo-lib-linux
/lib64/ld-linux-x86-64.so.2
```

That path does not exist on NixOS. It exists generally on other FHS-compatible Linux distributions.

To update this path, we use:

```
patchelf --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" ./typescript-demo-lib-linux
```

This will set the dynamic link loader to the relevant path (note this must be used inside nix-shell from a mkDerivation).

At this point we can now execute the `./typescript-demo-lib-linux`, however now it will tell us that a library is missing:

```
./typescript-demo-lib-linux: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory
```

We can tell by these 2 commands:

```
[nix-shell:~/Projects/Typescript-Demo-Lib/tmp]$ ldd ./typescript-demo-lib-linux	linux-vdso.so.1 (0x00007ffc3f9e1000)
	libdl.so.2 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/libdl.so.2 (0x00007f6f13796000)
	libstdc++.so.6 => not found
	libm.so.6 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/libm.so.6 (0x00007f6f13653000)
	libgcc_s.so.1 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/libgcc_s.so.1 (0x00007f6f13639000)
	libpthread.so.0 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/libpthread.so.0 (0x00007f6f13618000)
	libc.so.6 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/libc.so.6 (0x00007f6f13457000)
	/nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib/ld-linux-x86-64.so.2 => /nix/store/ikl21vjfq900ccbqg1xasp83kadw6q8y-glibc-2.32-46/lib64/ld-linux-x86-64.so.2 (0x00007f6f1379d000)
```

And also:

```
[nix-shell:~/Projects/Typescript-Demo-Lib/tmp]$ patchelf --print-needed ./typescript-demo-lib-linux
libdl.so.2
libstdc++.so.6
libm.so.6
libgcc_s.so.1
libpthread.so.0
libc.so.6
ld-linux-x86-64.so.2
```

The `libstdc++.so.6` is special shared library that needs to be acquired from `stdenv.cc.cc.lib`.

CMCDragonkai:
To get the path to `libstdc++.so.6` we need to do this:

```
»» ~
 ♜ nix repl                                                                                                                                       pts/4 23:02:28
Welcome to Nix version 2.3.10. Type :? for help.

nix-repl> :l <nixpkgs>
Added 14209 variables.

nix-repl> with pkgs; lib.makeLibraryPath [ stdenv.cc.cc.lib ]                                   
"/nix/store/5lgyy30nyvq59c4m2a7skmwaqz3jja3n-gcc-10.3.0-lib/lib"

nix-repl> 
```

I'm getting this path from my standard `<nixpkgs>` environment, which would not be the same as what would be inside the `nix-shell`. If we are inside nix-shell, a better way would be to acquire the `./pkgs.nix`:

```
[nix-shell:~/Projects/Typescript-Demo-Lib]$ nix repl
Welcome to Nix version 2.3.10. Type :? for help.

nix-repl> :l ./pkgs.nix
Added 14210 variables.

nix-repl> with pkgs; lib.makeLibraryPath [ stdenv.cc.cc.lib ] 
"/nix/store/5lgyy30nyvq59c4m2a7skmwaqz3jja3n-gcc-10.3.0-lib/lib"
```

Which turns out to be the same path coincidentally.

Now we can set the rpath which allows the dynamic link loader to look for the relevant dependency:

```
patchelf \
  --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" \
  --set-rpath /nix/store/5lgyy30nyvq59c4m2a7skmwaqz3jja3n-gcc-10.3.0-lib/lib \
  ./typescript-demo-lib-linux
```

The result is an executable that does execute, but has some JS error.

```
[nix-shell:~/Projects/Typescript-Demo-Lib/tmp]$ ./typescript-demo-lib-linux
pkg/prelude/bootstrap.js:1
ba7b810-9dad-11d1-80b4-00c04fd430c8",o.URL="6ba7b811-9dad-11d1-80b4-00c04fd430c8",o}("v5",80,(function(r){var e=[1518500249,1859775393,2400959708,3395469782],t=[1732584193,4023233417,2562383102,271733878,3285377520];if("string"==typeof r){var n=unescape(encodeURIComponent(r));r=[];for(var i=0;i<n.length;++i)r.push(n.charCodeAt(i))}else Array.isArray(r)||(r=Array.prototype.slice.call(r));r.push(128);for(var f=r.length/4+2,s=Math.ceil(f/16),u=new Array(s),c=0;c<s;++c){for(var l=new Uint32Array(16),p=0;p<16;++p)l[p]=r[64*c+4*p]<<24|r[64*c+4*p+1]<<16|r[64*c+4*p+2]<<8|r[64*c+4*p+3];u[c]=l}u[s-1][14]=8*(r.length-1)/Math.pow(2,32),u[s-1][14]=Math.floor(u[s-1][14]),u[s-1][15]=8*(r.length-1)&4294967295;for(var d=0;d<s;++d){for(var h=new Uint32Array(80),v=0;v<16;++v)h[v]=u[d][v];for(var y=16;y<80;++y)h[y]=o(h[y-3]^h[y-8]^h[y-14]^h[y-16],1);for(var g=t[0],b=t[1],w=t[2],U=t[3],A=t[4],I=0;I<80;++I){var m=Math.floor(I/20),C=o(g,5)+a(m,b,w,U)+A+e[m]+h[I]>>>0;A=U,U=w,w=o(b,30)>>>0,b=g,g=C}t[0]=t[0]+g>>>0,t[1]=t[1]+b>>>0,t[2]=t[2]+w>>>0,t[3]=t[3]+U>>>0,t[4]=t[4]+A>>>0}return[t[0]>>24&255,t[0]>>16&255,t[0]>>8&255,255&t[0],t[1]>>24&255,t[1]>>16&255,t[1]>>8&255,255&t[1],t[2]>>24&255,t[2]>>16&255,t[2]>>8&255,255&t[2],t[3]>>24&255,t[3]>>16&255,t[3]>>8&255,255&t[3],t[4]>>24&255,t[4]>>16&255,t[4]>>8&255,255&t[4]]}))}));{"mode":33188,"size":3267,"isFileValue":true,"isDirectoryValue":false,"isSocketValue":false,"isSymbolicLinkValue":false}import uuid from './dist/index.js';
        ^

SyntaxError: Invalid or unexpected token
    at new Script (vm.js:102:7)
    at readPrelude (internal/bootstrap/pkg.js:29:13)
    at internal/bootstrap/pkg.js:36:18
    at internal/bootstrap/pkg.js:43:4
    at internal/bootstrap/pkg.js:44:2
```

So the built executable isn't perfect, but it does sort of work. Given that we are now in JS error land, at least that means the runtime is running, but maybe using patchelf isn't great for executables built from vercel/pkg.

CMCDragonkai:
I have cloned `vercel/pkg` and did `npm install` and `npm run build`. This produces a `lib-es5` as the built output. And you run the `node ./lib-es5/bin.js --help` as the built `pkg`.

This will allow us to run `pkg` against our demo lib here, while also changing the code and seeing what exactly is causing the EPIPE error.

We know that it is happening at `child.stdin.write(h);`. The `child` is a `subprocess` object created by `spawn` function coming from the `child_process` library in Nodejs. The type is `ChildProcessByStdio`. The `child.stdin` is documented here: https://nodejs.org/api/child_process.html#child_process_subprocess_stdin

It appears the idea is to start a child process that is reading from STDIN, and then to write a `Buffer` of some bytes to this stdin. However we are getting an EPIPE error which usually means:

> “Broken pipe.” There is no process reading from the other end of a pipe. Every library function that returns this error code also generates a SIGPIPE signal; this signal terminates the program if not handled or blocked. Thus, your program will never actually see EPIPE unless it has handled or blocked SIGPIPE.

Which could mean a problem on piping, or a problem about the child process or a problem with how nodejs does this. But still need to test on Ubuntu to make sure this is actually a NixOS problem and not Linux in general.

CMCDragonkai:
The executable that is built on NixOS in nix-shell is actually a valid executable. In fact I think it's the actual node runtime just with no script being passed in. It's not something I can run on NixOS because the interpreter and rpath is set to the standard Linux paths, but after I copied it to WSL, I was able to run it, and and we get a javascript runtime error (which means the node runtime is working). This time it says that the `path` argument is undefined, which seems to mean it's the script path was never able to be written to this executable, and thus we didn't complete the bundling. This seems to make sense given that the EPIPE error is occuring where we are writing some sort of short string to the child process.

CMCDragonkai:
This reduces the error down to something involving the bundling of the actual JS source code with the Node runtime binary.

CMCDragonkai:
Finished build on Ubuntu. It's fine. So this is a problem specific to NixOS (maybe the nodejs build in NixOS).

CMCDragonkai:
If I disable bytecode generation, it works!

```
pkg . --targets linux --no-bytecode
```

Strangely I cannot refer to the exact file `pkg ./dist/bin/typescript-demo-lib.js --targets linux --no-bytecode` it throws this:

```
> Error! --no-bytecode and no source breaks final executable
  /home/cmcdragonkai/Projects/Typescript-Demo-Lib/dist/bin/typescript-demo-lib.js
  Please run with "-d" and without "--no-bytecode" first, and make
  sure that debug log does not contain "was included as bytecode".
```

The built executable I've copied to WSL Ubuntu and works.

CMCDragonkai:
This works now:

```
[nix-shell:~/Projects/Typescript-Demo-Lib]$ pkg . --no-bytecode
> pkg@5.3.0
> Targets not specified. Assuming:
  node14-linux-x64, node14-macos-x64, node14-win-x64
```

We get 3 executables.

CMCDragonkai:
Using `--no-bytecode` will be better for us since this is an open source project, so no need to obfuscate the code, and apparently it gives better reproducibility (but evidently not across different compilation platforms).

CMCDragonkai:
Realised that the reason might be the package.json... that's probably it.

---

Anyway, we can proceed!

CMCDragonkai:
But still strange error involving EPIPE - that is concerning if this is occurring only for NixOS. I suspect now this might be due to how nodejs is packaged in Nixpkgs. Perhaps the bytecode generation activates some feature that isn't available in the nodejs in Nixpkgs.

CMCDragonkai:
So I want to point out that unlike TypeScript-Demo-Electron, the name of this package is in the `package.json` which is:

```
@matrixai/typescript-demo-lib
```

That means the full name is org-scoped in NPM.

In our current default.nix we are using:

```
  name = "${builtins.replaceStrings ["/" "@"] ["_" ""] node2nixDev.packageName}-${node2nixDev.version}";
```

In order to remove `@` and replace `/` with `_`. That way the final name of the derivation here is:

```
matrixai_typescript-demo-lib-1.0.0
```

This is in line with other Nix packages which generally have the name with a tagged on version.

While this is the name of the derivation, we would not be using this as the name of the packaged executable.

In the TypeScript-Demo-Electron we instead have:

```
      name = "${utils.basename}-${version}-darwin-${arch}.zip";
```

It strips any `@matrixai/blah` to just `blah`. But appends the version and then the darwin plus architecture.

Our prereleases here should follow something similar:

```
name-of-package-version-platform-architecture[.ext]
```

That means we get something like:

```
typescript-demo-lib-1.0.0-win32-x64.exe
```

If we keep the org scoping we get:

```
matrixai_typescript-demo-lib-1.0.0-win32-x64.exe
```

I think we ignore the org scoping here since this isn't distributed on npm at this point, so we can use the proper package name.

CMCDragonkai:
GitHub claims to have a "latest" release: https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/linking-to-releases

That's probably because I don't have any releases technically, and I have only published tags, which they are all set to prereleases.

Instead a release creation is manual thing, while prereleases are automated.

CMCDragonkai:
To integrate pkg into the nix build pipeline, we will need to package up `nodePackages.pkg` similar to what we did with `@electron-forge/cli`.

CMCDragonkai:
We have a PR on nixpkgs upstream https://github.com/NixOS/nixpkgs/pull/128967

CMCDragonkai:
It was this commit that caused the change for bitwarden/cli: https://github.com/bitwarden/jslib/commit/1016bbfb9eb28c220de8d2ab86d1f2757328f254 It resulted in a new directory `jslib/common` which didn't exist before.

So I think node2nix is struggling to build bitwarden now.

If there's a way to bring vercel/pkg without attempting to update bitwarden, then this PR could be used, and somebody else has to fix `@bitwarden/cli`.

CMCDragonkai:
That PR is ready to be used. In order to be able to use it ourselves, we will need to wait for it to be merged upstream and then change our `pkgs.nix` to be pinned ahead. Adding its derivation to our overlay will be a bit of a pain since it's not just a simple derivation but can include references to other node packages. In the mean time we can rely on the npm installed pkg and just run the command within node2nix building environment...

CMCDragonkai:
I might need to change to using 3.1 binaries from pkg-fetch since pkg seems to fail on the mac build.

CMCDragonkai:
Relevant things:

* https://github.com/vercel/pkg#config - this is not needed for TypeScript-Demo-Lib (as there are no scripts/assets to load)
* https://github.com/vercel/pkg#snapshot-filesystem - this is important if we need to interact with the internal fs and deal with assets/scripts
* Auto-rewriting of the assets https://github.com/vercel/pkg#detecting-assets-in-source-code

One has to know that `scripts` are only needed for dynamic `require()` calls. This must however be normal `js` code (so no ts code allowed!)

The `assets` are for other things.

Still need to figure out what happens with:

* https://github.com/vercel/pkg#native-addons - that will be for the next PR to test how it deals with things like utp-native

@scottmmorris @DrFacepalm

CMCDragonkai:
It's all working!

CMCDragonkai:
We get:

```
/nix/store/j4vbjipszxxynvvv466jlzq9qmkqvq2r-typescript-demo-lib-1.0.0-macos-x64
```

CMCDragonkai:
It's time to test gitlab CI/CD now for the prereleases.

PR #26, opened 2021-07-02
Change from darwin to macos

CMCDragonkai:
It's silly to call what we are building darwin, it's really macos.

Also matches behaviour with https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/25


CMCDragonkai:
Woops. This is wrong.

PR #29, opened 2021-07-21
Pkg integration

### LINKED ISSUE #27: Integrate a Node native module like utp-native or leveljs in order to demonstrate vercel/pkg
The vercel/pkg has successfully been implemented and is now producing pre-release packages in our release page.

However we have yet to test how it works when dealing with native modules like utp-native.

We need to add in utp-native and leveljs as demonstrations of building a CLI executable with native code.

This will be essential for js-polykey to be capable of producing release executables.

1. Integrate leveldb into TypeScript-Demo-Lib
2. Integrate utp-native into TypeScript-Demo-Lib
3. Perform a hello world equivalent of using these 2 libraries, for leveldb that's CRUD on the database, for utp-native we would need to start a server, and close the server.
4. Then use vercel/pkg to package up the program
5. Sanity check the release
6. Fix up any native module configuration required for vercel/pkg

### Checklist:
- [x] I've done some reading on native modules and vercel/pkg.
- [x] cloned the TypeScript-Demo-Lib, built and ran it to see it working.
- [x] built TypeScript-Demo-Lib with pkg and ran it in linux to check it is working.
- [x] Written a hello world test for level.
- [x] written a hello world test for utp-native.
- [x] pkg with level and test it works.
- [x] pkg with utp-native and test it works.
- [x] create a test for worker threads.
- [x] pkg with worker threads and test it works

---

Relevant upstream PR https://github.com/vercel/pkg/pull/1273

tegefaulkes:
### Description
This PR is to test using vercel/pkg with native modules.
It imports the following node modules:
- level
- utp-native
- threads.js

It should be able to build via pkg and nix-build method.

### Issues Fixed
* Fixes #27 

### Tasks
- import, build and test level
- import, build and test utp-native
- import, build and test threads

### Final checklist
- import and create hello world example for:
    - [x] level
    - [x] utp-native
    - [x] threads
- build and test via pkg
    - [x] level
    - [x] utp-native
    - [x] threads
- build and test via nix-build
    - [x] level
    - [x] utp-native
    - [x] threads
- [x] modify release.nix to include the required prebuilds in the result.
- [x] final tests of the nix builds. see if they include prebuilds properly.
- [x] test if GH auto release still works.

---

Related issues and PRs:

* https://github.com/vercel/pkg/issues/1264
* https://github.com/vercel/pkg/pull/1273

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,13 +73,13 @@ $ npm run ts-node
 
 You can also create test files in `./src`, and run them with `npm run ts-node ./src/test.ts`.
 
-This allows you to test individual pieces of typescript code and it makes it easier when doing large scale rearchitecting of TypeScript code.
+This allows you to test individual pieces of typescript code, and it makes it easier when doing large scale architecting of TypeScript code.
 
 ### Path Aliases
 
 Due to https://github.com/microsoft/TypeScript/issues/10866, you cannot use path aliases without a bundler like Webpack to further transform the generated JavaScript code in order to resolve the path aliases. Because this is a simple library demonstration, there's no need to use a bundler. In fact, for such libraries, it is far more efficient to not bundle the code.
 
-However we have left the path alias configuration in `tsconfig.json`, `jest.config.js` and in the tests we are making use of the `@` alias.
+However, we have left the path alias configuration in `tsconfig.json`, `jest.config.js` and in the tests we are making use of the `@` alias.
 
 #### VSCode Path Aliases
 
@@ -98,6 +98,70 @@ To resolve this, add `./tests/**/*` into the `tsconfig` `include` section:
 This will however make `tsc` build the `tests` into the `dist` output.
 
 **Therefore this fix should only be done in your own workspace, do not commit or push this change up.**
+
+### Native Module Toolchain
+
+There are some nuances when packaging with native modules.
+Included native modules are level witch include leveldown and utp-native.
+
+If a module is not set to public then pkg defaults to including it as bytecode.
+To avoid this breaking with the `--no-bytecode` flag we need to add `--public-packages "*"`
+
+#### leveldown
+
+To get leveldown to work with pkg we need to include the prebuilds with the executable.
+after building with pkg you need to copy from `node_modules/leveldown/prebuilds` -> `path_to_executable/prebuilds`
+You only need to include the prebuilds for the arch you are targeting. e.g. for linux-x64 you need `prebuild/linux-x64`.
+
+The folder structure for the executable should look like this.
+- linux_executable_elf
+- prebuilds
+  - linux-x64
+    - (node files)
+
+#### utp-native
+
+Including utp-native is simpler, you just need to add it as an asset for pkg.
+Add the following lines to the package.json.
+```json
+"pkg": {
+    "assets": "node_modules/utp-native/**/*"
+  }
+```
+
+#### threads.js
+
+To make sure that the worker threads work properly you need to include the compiled worker scripts as an asset.
+This can be fixed by adding the following to `package.json`
+
+```json
+"pkg": {
+    "assets": "dist/bin/worker.js"
+  }
+```
+
+If you need to include multiple assets then add them as an array.
+
+```json
+"pkg": {
+    "assets": [
+      "node_modules/utp-native/**/*",
+      "dist/bin/worker.js"
+    ]
+  }
+```
+
+#### Integration into Nix
+
+Nix build uses node2nix to create the dependencies of the node modules. this does a fairly good job of detecting dependencies but with native modules it may fail to detect CLI tools like `node-gyp-build`.
+
+To ensure the proper dependencies exist while building we need to bring in these utilities during the build:
+
+```
+buildInputs = attrs.buildInputs ++ [ nodePackages.node-gyp-build ];
+```
+
+This has to be done to both `node2nixProd` and `node2nixDev`.
 
 ### Docs Generation
 
--- a/pkgs.nix
+++ b/pkgs.nix
@@ -1,5 +1,4 @@
 import (
-  # let rev = "60cce7e5e1fdf62421ef6d4184ee399b46209366"; in
-  let rev = "04a48a2fcd2370f4e8a8c9fd3394531ccdb30f1e"; in
+  let rev = "53caacaf56640d04180775aee016d2f16d6f083c"; in
   fetchTarball "https://github.com/MatrixAI/nixpkgs/archive/${rev}.tar.gz"
 )
--- a/release.nix
+++ b/release.nix
@@ -2,11 +2,6 @@
 
 with pkgs;
 let
-  pkg = pkgs.nodePackages.pkg.override {
-    postFixup = ''
-      patch -p0 < ${./nix/leveldown.patch}
-    '';
-  };
   utils = callPackage ./utils.nix {};
   buildElf = arch:
     stdenv.mkDerivation rec {
@@ -14,7 +9,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        pkg
+        utils.pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
@@ -40,7 +35,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        pkg
+        utils.pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
@@ -65,7 +60,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        pkg
+        utils.pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
--- a/release.nix
+++ b/release.nix
@@ -18,6 +18,8 @@ let
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
+      # ensure that native modules are built from source
+      npm_config_build_from_source = "true";
       buildPhase = ''
         cp ${./package.json} package.json
         pkg . \
@@ -42,6 +44,8 @@ let
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
+      # ensure that native modules are built from source
+      npm_config_build_from_source = "true";
       buildPhase = ''
         cp ${./package.json} package.json
         pkg . \
@@ -65,6 +69,8 @@ let
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
+      # ensure that native modules are built from source
+      npm_config_build_from_source = "true";
       buildPhase = ''
         cp ${./package.json} package.json
         pkg . \
--- a/release.nix
+++ b/release.nix
@@ -2,6 +2,11 @@
 
 with pkgs;
 let
+  pkg = pkgs.nodePackages.pkg.override {
+    postFixup = ''
+      patch -p0 < ${./nix/leveldown.patch}
+    '';
+  };
   utils = callPackage ./utils.nix {};
   buildElf = arch:
     stdenv.mkDerivation rec {
@@ -9,7 +14,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        nodePackages.pkg
+         pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
@@ -18,7 +23,9 @@ let
         pkg . \
           --targets linux-${arch} \
           --no-bytecode \
-          --output out
+          --public-packages "*" \
+          --output out\
+          --verbose
       '';
       installPhase = ''
         cp out $out
@@ -31,7 +38,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        nodePackages.pkg
+        pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
@@ -40,6 +47,7 @@ let
         pkg . \
           --targets win-${arch} \
           --no-bytecode \
+          --public-packages "*" \
           --output out.exe
       '';
       installPhase = ''
@@ -53,7 +61,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-        nodePackages.pkg
+        pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
@@ -62,6 +70,7 @@ let
         pkg . \
           --targets macos-${arch} \
           --no-bytecode \
+          --public-packages "*" \
           --output out
       '';
       installPhase = ''
--- a/release.nix
+++ b/release.nix
@@ -14,7 +14,7 @@ let
       version = utils.node2nixDev.version;
       src = "${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}";
       buildInputs = [
-         pkg
+        pkg
       ];
       PKG_CACHE_PATH = utils.pkgCachePath;
       PKG_IGNORE_TAG = 1;
--- a/src/bin/typescript-demo-lib.ts
+++ b/src/bin/typescript-demo-lib.ts
@@ -10,9 +10,8 @@ import testUptNative from '../lib/test-utp-native';
 
 async function main(argv = process.argv): Promise<number> {
   // Print out command-line arguments
-  const argArray = argv.slice(2);
-  const args = argArray.toString();
-  process.stdout.write('[' + args + ']\n');
+  argv = argv.slice(2); // Removing prepended file paths.
+  process.stdout.write('[' + argv.slice(0, 2).toString() + ']\n');
 
   // Create a new Library with the value someParam = 'new library'
   // And print it out
@@ -23,12 +22,13 @@ async function main(argv = process.argv): Promise<number> {
   process.stdout.write(uuidv4() + '\n');
 
   // Add the first two command-line args and print the result
-  const nums = new NumPair(parseInt(argArray[0]), parseInt(argArray[1]));
+  const nums = new NumPair(parseInt(argv[0]), parseInt(argv[1]));
   const sum = nums.num1 + nums.num2;
   process.stdout.write(nums.num1 + ' + ' + nums.num2 + ' = ' + sum + '\n');
 
   // Testing native modules.
-  await testLevel(argv[0]);
+  const dir = argv[2] ?? '.';
+  await testLevel(dir);
   await testWorkers();
   await testUptNative();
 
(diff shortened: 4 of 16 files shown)


src/bin/worker.ts
```diff
@@ -0,0 +1,7 @@
+import { expose } from "threads/worker"
```
CMCDragonkai, line 1: This should got to `src/worker.ts` or `src/workers/worker.ts` like we do in `js-encryptedfs` or `js-polykey`.

utils.nix
```diff
@@ -28,10 +28,12 @@ rec {
   node2nixProd = (
     (import (node2nixDrv false) { inherit pkgs nodejs; }).shell.override {
       dontNpmInstall = true;
+      buildInputs = [ nodePackages.node-gyp-build ];
     }
   ).nodeDependencies;
   node2nixDev = (import (node2nixDrv true) { inherit pkgs nodejs; }).package.overrideAttrs (attrs: {
     src = src;
+    buildInputs = attrs.buildInputs ++ [ nodePackages.node-gyp-build ];
```
CMCDragonkai, line 36: Have a check whether both require this override, or just one of `node2nixProd` and `node2nixDev`.

tegefaulkes:
Completed most of the checklist, The last thing to do is check of the GH releases works properly with the changes.
Depending on the output we may have to zip the builds now since they are directories instead of just the executable.

src/bin/typescript-demo-lib.ts
```diff
@@ -24,6 +27,50 @@ function main(argv = process.argv): number {
   const sum = nums.num1 + nums.num2;
   process.stdout.write(nums.num1 + ' + ' + nums.num2 + ' = ' + sum + '\n');
 
+  // Level tests.
```
CMCDragonkai, line 30: Don't use `console.log` here. It makes the test harder.

Use `process.stdout.write` like above.

Also I'd say put these into `src/lib/...` for each individual functionality. Provider functions like `level.ts` which exports a function that does the below, and same for the other ones.

release.nix
```diff
@@ -59,13 +65,16 @@ let
       PKG_IGNORE_TAG = 1;
       buildPhase = ''
         cp ${./package.json} package.json
+        mkdir -p output/prebuilds
```
CMCDragonkai, line 68: Does this need `output/prebuilds`? Why not just `./prebuilds`? The `./` is the working directory of the build. As long as this directory doesn't conflict with anything in the `./` which I believe is the produced source files from the upstream derivations.

Remember this is an overridden derivation, the `$src` has already unpacked into the current directory.

release.nix
```diff
@@ -59,13 +65,16 @@ let
...
+        cp -r node_modules/leveldown/prebuilds/darwin-${arch} output/prebuilds/darwin-${arch}
         pkg . \
           --targets macos-${arch} \
           --no-bytecode \
-          --output out
+          --public-packages "*" \
+          --output output/TypeScript-Demo-Lib
```
CMCDragonkai, line 74: Why are we creating a `output` directory?

The `$out` is obviously for the final executable. But I see you're putting `prebuilds` into the `output`. Is that not possible with just the `./` directory?

release.nix
```diff
@@ -15,19 +15,22 @@ let
...
       installPhase = ''
-        cp out $out
+        cp -r output $out
       '';
       dontFixup = true;
     };
   buildExe = arch:
     stdenv.mkDerivation rec {
-      name = "${utils.basename}-${version}-win32-${arch}.exe";
+      name = "${utils.basename}-${version}-win32-${arch}";
```
CMCDragonkai, line 33: Why change this? The name of the windows executable has an extension of `.exe`.

release.nix
tegefaulkes: When I use `./` instead of a temp directory, the result included all of the src files.
I also feel that creating and using a temp directory makes the process more explicit.

CMCDragonkai:
Is there a way for you test the CI/CD execution of your changes.

You just need to go into the gitlab yaml file and enable this for your branch as well.

Then go to https://gitlab.com/MatrixAI/open-source/typescript-demo-lib/-/pipelines to manually trigger it. Or https://gitlab.com/MatrixAI/open-source/typescript-demo-lib/-/jobs.

release.nix
tegefaulkes: The output is a directory now that includes the .exe

release.nix
CMCDragonkai: Is that because of:

```
--public-packages "*" \
```

Is the `*` picking up everything?

CMCDragonkai: If you must use a a special directory for this. I'd say use `pkg` as the special directory for `pkg` related building.

release.nix
CMCDragonkai: You should leave it as the actual executable. Is there other files necessary for the actual `$out`?

This is not a nix package, so the executable as `$out` is totally fine.

Plus this actually is integrated into the Gitlab CI/CD which expects a single file as `$out`. Changing it to a directory adds extra work to do there.

README.md
```diff
@@ -117,3 +117,75 @@ npm publish --access public
 git push
 git push --tags
 ```
+
+
```
CMCDragonkai, line 121: Remove empty line here.

README.md
```diff
@@ -117,3 +117,75 @@ npm publish --access public
...
+### Packaging with vercel.pkg
```
CMCDragonkai, line 122: Change this to `### Releasing`.

And fix up the typos in the text. Put it into word/writer if necessary.

CMCDragonkai: Spacing of headers should be:

```
text text test

### Header

text text text
```

README.md
```diff
@@ -117,3 +117,75 @@ npm publish --access public
...
+```json
+"pkg": {
+    "assets": [
+      "node_modules/utp-native/**/*",
+      "dist/bin/worker.js"
+    ]
+  }
+```
+
+#### working with nix-build
```
CMCDragonkai, line 169: This should be: `### Native Module Toolchain`.

This section actually not part of `Releasing`.

It should be it's own larger section that comes before `Docs Generation`.

README.md
```diff
@@ -117,3 +117,75 @@ npm publish --access public
...
+You only need to include the prebuild for the arch you are targeting. eg for linux-x64 you need `prebuild/linux-x64`.
+
+The folder structure for the executable should look like this.
+- linux_executable_elf
+- prebuilds
+  - linux-x64
+    - (node files)
+
+#### utp-native
+Including utp-native is simpler, you just need to add it as an asset for pkg.
```
CMCDragonkai, line 141: Can you explain why `utp-native` is different from `leveldown`?

pkgs.nix
```diff
@@ -1,5 +1,6 @@
 import (
   # let rev = "60cce7e5e1fdf62421ef6d4184ee399b46209366"; in
-  let rev = "04a48a2fcd2370f4e8a8c9fd3394531ccdb30f1e"; in
+  # let rev = "04a48a2fcd2370f4e8a8c9fd3394531ccdb30f1e"; in
+  let rev = "53caacaf56640d04180775aee016d2f16d6f083c"; in
```
CMCDragonkai, line 4: Which revision is this again? Are we updating this to the latest?

CMCDragonkai:
> Is there a way for you test the CI/CD execution of your changes.
> 
> You just need to go into the gitlab yaml file and enable this for your branch as well.
> 
> Then go to https://gitlab.com/MatrixAI/open-source/typescript-demo-lib/-/pipelines to manually trigger it. Or https://gitlab.com/MatrixAI/open-source/typescript-demo-lib/-/jobs.

Oh if you do this, make sure to switch it back later, we only want builds to occur on master, not on every branch.

src/bin/worker.ts
tegefaulkes: moved to `src/lib/workers/worker.ts` and moved the workers test into `src/lib/workers/test-workers.ts`

src/bin/typescript-demo-lib.ts
tegefaulkes: - Replaced each use of `consoel.log()` with `process.stdout.write()`.
- Separated each tests into it's own file inside `src/lib/`

release.nix
tegefaulkes: Ok, I can change it to pkg.
as for `--public-packages "*" \` is just flagging everything is a public package since non public packages are included as bytecode by default.. This breaks with the `--no-bytecode` flag.

release.nix
tegefaulkes: - changing `output` to `pkg`. relevant to above comment.

CMCDragonkai:
We need `--no-bytecode` because without this flag it doesn't work inside 
Nix and it isn't deterministic. So up to your decision. As I said if you 
really need a directory, try `pkg` as the directory name to avoid confusion.

release.nix
tegefaulkes: We need to have the output as a directory since we need to include the prebuilds for leveldown.
we can have it as a zip file if needed.

README.md
tegefaulkes: I'm not sure, they include the native bindings almost identically.
leveldown `binding.js`
```
module.exports = require('node-gyp-build')(__dirname)
```
where as utp-native does.
`lib/binding.js`
```
const path = require('path')

module.exports = require('node-gyp-build')(path.join(__dirname, '..'))

```

utils.nix
tegefaulkes: Just having it in node2nixProd doesn't work while just having it in node2nixDev does.
Removing it from node2nixProd.

release.nix
tegefaulkes: Using pkg 5.3.1 doesn't fix the issue for leveldown prebuilds, neither does adding them as assets.
this seems to be an issue with pkg itself, I've posted an issue at https://github.com/vercel/pkg/issues/1264.

pkgs.nix
tegefaulkes: I picked the latest pkgs commit when I started on this issue.

CMCDragonkai:
You mean nixpkgs commit from master? Does this all work with just the 
revision provided by nixpkgs? 
https://github.com/NixOS/nixpkgs/commit/3d64b6067f1c457f889e91c59ced273be1d1e4f1

The above is the just the one that brought in `pkg`.

tegefaulkes:
I have created an upstram PR addressing the issue with the leveldown warnings.
vercel/pkg#1273

tegefaulkes:
When this is done, the changes for js-polykey are not far behind. the only complication would be the jose/jwt throwing warnings. But I've added them as an asset and it seems to be working with manual testing e.g. starting agent + starting a session.

CMCDragonkai:
I'm checking out your change to `vercel/pkg` now.

Firstly:

```
git fetch origin pull/1273/head:pr-1273
git checkout pr-1273
# setup nix-shell here
npm install
npm run build
# the build outputs to ./lib-es5, however the executable requires the node interpreter
node ./lib-es5/bin.js  --help
```

The output:

```

  pkg [options] <input>

  Options:

    -h, --help           output usage information
    -v, --version        output pkg version
    -t, --targets        comma-separated list of targets (see examples)
    -c, --config         package.json or any json file with top-level config
    --options            bake v8 options into executable to run with them on
    -o, --output         output file name or template for several files
    --out-path           path to save output one or more executables
    -d, --debug          show more information during packaging process [off]
    -b, --build          don't download prebuilt base binaries, build them
    --public             speed up and disclose the sources of top-level project
    --public-packages    force specified packages to be considered public
    --no-bytecode        skip bytecode generation and include source files as plain js
    -C, --compress       [default=None] compression algorithm = Brotli or GZip

  Examples:

  – Makes executables for Linux, macOS and Windows
    $ pkg index.js
  – Takes package.json from cwd and follows 'bin' entry
    $ pkg .
  – Makes executable for particular target machine
    $ pkg -t node14-win-arm64 index.js
  – Makes executables for target machines of your choice
    $ pkg -t node12-linux,node14-linux,node14-win index.js
  – Bakes '--expose-gc' and '--max-heap-size=34' into executable
    $ pkg --options "expose-gc,max-heap-size=34" index.js
  – Consider packageA and packageB to be public
    $ pkg --public-packages "packageA,packageB" index.js
  – Consider all packages to be public
    $ pkg --public-packages "*" index.js
  – Bakes '--expose-gc' into executable
    $ pkg --options expose-gc index.js
  – reduce size of the data packed inside the executable with GZip
    $ pkg --compress GZip index.js
```

I'm going to assume you manage to make it work.. so he main thing is that we need to get node2nix to work against your fork's branch.

CMCDragonkai:
@tegefaulkes In the future for this sort of work you would want to create the forks under MatrixAI group in github rather than your personal account: https://github.com/tegefaulkes/pkg

It makes it easier for us to maintain forks and maintain overrides in our nixpkgs overlay.

CMCDragonkai:
The other really important thing is to always branch off from `main` or `master` when creating a PR. This way you can have a `main` that syncs up with upstream, while creating a feature branch to be PRed into upstream main.

So for now I'll just create a feature branch off `main` and do the same as what you did.

CMCDragonkai:
This is now done here: https://github.com/MatrixAI/pkg/tree/leveldown

The steps were:

```
git remote add matrixai git@github.com:MatrixAI/pkg.git
git checkout main
git checkout -b leveldown
# make the change
git add -A
git commit -m 'Removes dictionary override for bundling leveldown - Fixes vercel#1264'
git push matrixai
```

CMCDragonkai:
The nix override work in our overlay is located here: https://gitlab.com/MatrixAI/Engineering/nixpkgs-overlay/-/merge_requests/15

pkgs.nix
CMCDragonkai: You should usually get the commit hash matching the exact merge commit. Which would be this one: https://github.com/NixOS/nixpkgs/commit/3d64b6067f1c457f889e91c59ced273be1d1e4f1

Notice you can see it here:

image

CMCDragonkai: That's for in the future, which avoids bringing in other confounding variables/changes.

tegefaulkes: Ah, I'll keep that in mind.

tegefaulkes:
Ok, should be working now. Last thing is to check if the github CI still works.

pkgs.nix
```diff
@@ -1,5 +1,13 @@
...
+#  # let rev = "60cce7e5e1fdf62421ef6d4184ee399b46209366"; in
+#  # let rev = "04a48a2fcd2370f4e8a8c9fd3394531ccdb30f1e"; in
+#  let rev = "53caacaf56640d04180775aee016d2f16d6f083c"; in
+#  fetchTarball "https://github.com/MatrixAI/nixpkgs/archive/${rev}.tar.gz"
+#)
+
+# Pinning to an overlay with a patch for vercel/pkg
+import (builtins.fetchGit {
+  url = "git@gitlab.com:MatrixAI/Engineering/nixpkgs-overlay.git";
+  rev = "958ab694978371a8d73c328cb515df85c7150960";
```
CMCDragonkai, line 11: No need for this, just use the branch name below is sufficient. Plus I've rebased and squashed on the branch.

pkgs.nix
```diff
@@ -1,5 +1,13 @@
-import (
-  # let rev = "60cce7e5e1fdf62421ef6d4184ee399b46209366"; in
-  let rev = "04a48a2fcd2370f4e8a8c9fd3394531ccdb30f1e"; in
-  fetchTarball "https://github.com/MatrixAI/nixpkgs/archive/${rev}.tar.gz"
-)
+#import (
```
CMCDragonkai, line 1: Remove these comments and just write a comment on this PR about how you came to using the relevant commits.

CMCDragonkai:
I don't think anything needs to change in our `.gitlab-ci.yml` file. It should work fine now.

So please fix up my comments, ensure any linting is done, and also clean up the README and format it and squash all the commits.

CMCDragonkai:
See comment here https://gitlab.com/MatrixAI/Engineering/nixpkgs-overlay/-/merge_requests/15/diffs#note_642632986 regarding our policy of using our private overlay in public packages.

tegefaulkes:
K, fixing up now.

release.nix
CMCDragonkai: Can this be resolved now back to the way it was?

CMCDragonkai:
After everything, don't forget to test out the Gitlab CI/CD with this method: https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/29#issuecomment-884637611

Then finally we can merge.

tegefaulkes:
I switched back to the public pkgs and added in the patch.
I've tested it in wsl and windows. It seems to be working. I'll try to run a pipeline in a second.

tegefaulkes:
Misclicked closed. My bad.

tegefaulkes:
Failing on tests. I'll look it over and fix it up tomorrow.

tegefaulkes:
Checks are passing, looks like it's working. this should be good for a merge after a review. @CMCDragonkai

tegefaulkes:
Squashed.

tegefaulkes:
That should be the last of it.
- [x] rebase
- [x] small fixes
- [x] squash
- [ ] review
- [ ] merge

CMCDragonkai:
Just fixed up a few things:

* Removed useless comments from pkgs.nix
* Formatted README.md with spacing between the code blocks and text
* Restoring the mocks should happen at the end of each test
* Indentation issues in `release.nix`

CMCDragonkai:
I noticed you also enabled source map for this repo. This is a good idea for a library. For a end-user application, not sure if it is relevant.

tegefaulkes:
> I noticed you also enabled source map for this repo. This is a good idea for a library. For a end-user application, not sure if it is relevant.

Ah, jest was throwing some minor errors without it. It's not needed I think.

CMCDragonkai:
You should also always test: `nix-build ./release.nix --attr application`.

Also we now have a custom override for our `pkg` here, this would have to done for all the repos using pkg.

Until upstream merges. Our overlay can also have this change applied too in the mean time.

CMCDragonkai:
The application build is currently failing due to the missing `node-gyp-build` command. It needs to be available to the derivation. We've dealt with this before.

CMCDragonkai:
So it turns out that what's required is that:

```
     node2nixProd = (
  -    (import (node2nixDrv false) { inherit pkgs nodejs; }).shell.override {
  +    (import (node2nixDrv false) { inherit pkgs nodejs; }).shell.override (attrs: {
  +      buildInputs = attrs.buildInputs ++ [ nodePackages.node-gyp-build ];
         dontNpmInstall = true;
  -    }
  +    })
```

The `override` can also take a function which takes the `attrs`, originally this was not the case. And there was a difference between `override` and `overrideAttrs`.

Interestingly you cannot make this work if you use `overrideAttrs`.

CMCDragonkai:
Now I find that you need `node-gyp-build` for both `node2nixProd` and `node2nixDev`.

This makes sense as the `node2nixProd` will only install production dependencies, whereas `node2nixDev` will install both production and development dependencies. So if there's an override, both require the `node-gyp-build` to be available.

Now that override and `overrideAttrs` both take a function with `attrs`, I wonder if this is because `override` delegates to `overrideAttrs` if it's an attribute function. Because nixpkgs upstream says that `override` is meant to override the function parameters. But this doesn't make any sense if there is access to existing attributes.

CMCDragonkai:
Yea I checked `override` being able to take `attrs` and applying something like `buildInputs` is specific to node2nix.

CMCDragonkai:
Regarding override issues: https://github.com/svanderburg/node2nix/issues/178

I've changed to using `override` instead of `overrideAttrs` for both `node2nixDev` and `node2nixProd`.

Regarding the build warnings of creating a homeless shelter that's cause `pkg` is attempting to download prebuilt-binaries but instead building from source should be possible.

This is solved with this environment variable in the release builds:

```
      # ensure that native modules are built from source
      npm_config_build_from_source = "true";
```

Regarding the application build, both dev and prod require the `node-gyp-build` due to leveldown. So this will be important for any future native modules we are adding in, the build system will become ever more complex. So it's ideal we document what each thing is for.

---

@tegefaulkes still has to fix up the executable itself, the arguments to `./result/bin/typescript-demo-lib` should be all optional. I think some arguments were introduced by @emmacasolin before where there was no defaulting value for them. Right now not passing it the right set of args results in weird errors.

This is also important for the docker output which doesn't pass any parameters to the command. It will look like something is broken without this being fixed.

CMCDragonkai:
My last commit ensures parity between nix-shell and nix-build. This should ensure that `pkg` being used inside `nix-shell` is the same as that which is used for the release. These changes will need to be applied to js-polykey too.

At the final point make sure to test all builds, application, docker, and the executables. And then merging time!

PR #31, opened 2021-08-16
Changed to using scripts for polykeyWorker.js

CMCDragonkai:
### Description

According to @tegefaulkes this is the correct way of setting `polykeyWorker.js`. I've made the change here. Can you check if windows/linux builds work? If so, merge this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
   "pkg": {
     "assets": [
       "node_modules/leveldown/**/*",
-      "node_modules/utp-native/**/*",
-      "dist/lib/workers/worker.js"
-    ]
+      "node_modules/utp-native/**/*"
+    ],
+    "scripts": "dist/lib/workers/worker.js"
   },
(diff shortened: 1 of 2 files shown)


tegefaulkes:
Runs fine in WSL. But I think a change in the master is messing up the `.exe`.
Level is failing to create the database file since the path is invalid. I think you changed the path for that to use `/tmp/` at some point?

tegefaulkes:
Yeah, That absolute path is not valid in windows.
```diff
- const dir = argv[2] ?? '.';
+  const dir = argv[2] ?? '/tmp';
  await testLevel(dir);
  ```

CMCDragonkai:
Yea I forgot about that. Instead of using `/tmp`. Use https://nodejs.org/api/os.html#os_os_tmpdir instead. Please make the change and merge.

tegefaulkes:
On it.

Issue #32, opened 2021-10-07
Suport Dynamic Imports and Nodejs Module Type


CMCDragonkai:
### What is your research hypothesis/question?

There is a conjunction of 4 features:

* dynamic imports
* top level await
* nodejs module type (es modules in nodejs)
* explicit exports in package.json

That enables some nifty capabilities:

* The ability to create modules that can do feature tests and import the right module for a given platform - cross platform compatibility (especially in terms of platform-native features vs importing specific modules)
* The ability to move away from commonjs modules and use esmodules as a standard for all platforms - this can reduce the dependence on bundling if all platforms support esmodules
* The ability to create explicit package exports, so one can create subdirectory imports `import x from 'a/b/c';` - this can allow us to flexibly export things without having to go down the `dist/...` route all the time

These features are all coming online in various forms, however some initial experiments show that they involve alot of changes.

### Review existing ideas, literature and prior work

1. [ ] - motivation from nativescript https://github.com/MatrixAI/js-polykey/issues/155#issuecomment-935348551
2. [ ] - experiment in js-id https://github.com/MatrixAI/js-id/issues/1#issuecomment-936105951
3. [ ] - ts-node support https://github.com/TypeStrong/ts-node/issues/1007

### Sub-Issues & Sub-PRs created

1. ...
2. ...
3. ...

CMCDragonkai:
It would be nice if changing to this didn't involve changing all of our import paths.

CMCDragonkai:
Pkg doesn't currently support ES modules. But it should soon: https://github.com/vercel/pkg/issues/782.

Currently it may require us to use `esbuild`.

The `esbuild` could be trialed out to see if it's better than using webpack which we had removed before, and see if it can help also deal with non-transpiled files that are in `./src` and automatically move to `./dist`.

It may also help with doing some optimisation work like tree-shaking. But of course it doesn't make sense to put it all into one single file like we were doing in webpack. Library modules should be able to imported separately.

CMCDragonkai:
Note that dynamic imports already work. Right now they are just translated to `require` calls. However being able to use these for cross-platform implementation swapping will require top level awaits.

CMCDragonkai:
With #37 merged, we are one step closer.

Before working with ES modules and dynamic imports that use ES modules, it must be tested that it works in vercel pkg. Keep track of this issue: https://github.com/vercel/pkg/issues/1603

CMCDragonkai:
[...]

PR #33, opened 2021-11-03
Enable requiring the package.json

CMCDragonkai:
### Description

The `package.json` is a useful to be able to read inside the source code. This is because it contains version information that we often need to provide in our own program.

There are 2 ways of acquiring the `package.json`:

```
import packageJson from '../package.json';
const packageJson = require('../package.json');
```

The former is using ES6 imports, and will trigger typescript to resolve and compile it.

TSC will ask for `resolveJsonModule: true` option in `tsconfig.json`. However if we enable it to be true, then the `package.json` file will be copied in the `dist/` directory.

```
dist
├── package.json
└── src
```

At this point in time, we don't really want this, although we could adapt by changing the `main` and `types` path in `package.json`.

Ideally TSC will just not copy `package.json`, but there's only 2 ways to do this.

First way is to remove `resolveJsonModule: true` so that it's not resolved. Then make tsc ignore the error:

```ts
// @ts-expect-error package.json is outside the compiled dist
import { version } from '../package.json';
```

Then when compiling it works.

Second way, just use `require`, and tsc will ignore it.

The second way is not as good idea if we want to later move to ES6 style imports/exports as in #32.

The first way has one limitation, the `resolveJsonModule` must be false.  This means we are making TSC not check/resolve the json module at all and therefore there's no type information. This means we cannot use import json module anywhere, and TSC won't have any type information about JSON that is imported.

This can be a problem, and then JSON would have to be considered non-source code, but instead static assets that should be acquired elsewhere.

**An alternative to this solution is https://stackoverflow.com/a/61467483/582917 and https://stackoverflow.com/a/61467483/582917. These suggest creating a `tsconfig.json` under `src` as well as `tsconfig.json` in the root.**

In our PK code, we have to use `resolveJsonModule: true` (due to our json schema), so our only option is to use `require` right now. We must continue this PR to try the other solutions so that #32 can be used.

### Tasks

1. [x] - Try out the `require` method
2. [x] - Try out `@ts-ignore` method
3. [ ] - Try out the super and subproject method

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
A dynamic import has the same problem, if `--resolveJsonModule` is used it ends up acquiring it and copying it. But in that case no need for `@ts-expect-error`.

CMCDragonkai:
Dynamic imports would require top level await. And that requires using ES6 modules. So for now that's not an option until we try #32.

But I found a solution to this problem.

The idea is to use `rootDir` to say `./src`. This forces TSC to always treat `src` as the root dir. Then when importing from `../package.json` it will cause an error. This is when we use `@ts-ignore` to ignore it.

Strangely, the `@ts-expect-error` cannot be used, I guess it's in a different part of the pipeline. But doing this allows us to enable `resolveJsonModule` and then we can in fact import JSON from inside `./src` and `../package.json` is just a special case.

CMCDragonkai:
Ok I think this will work in PK, merging.

PR #37, opened 2022-04-18
Upgrading to node 16.x LTS (and pkgs.nix upgrade to 21.11 or later)

### LINKED ISSUE #32: Suport Dynamic Imports and Nodejs Module Type

CMCDragonkai:
### Description

Node 14 is already EOL. We should be upgrading to node 16.x LTS. This introduces some major features that we want in NodeJS:

* #32
* https://github.com/MatrixAI/js-id/issues/2
* https://github.com/MatrixAI/js-polykey/issues/297

This includes some other changes:

* Updating `pkgs.nix` to 21.11 (or later/master) which has `nodejs-16_x` but `nodejs` is still `14.19.1` (we may need to go to master instead
* Using `node2nix` at `1.10.0` which supports `16.x`, currently `1.9.0` does not support `16.x`
* Updating to `pkg` to `5.6.0`, which requires an update to our `pkgBuilds` in `utils.nix`. As it would no longer be using `3.1` and no longer using `14.17.0` and upgrade to v16
* Fix typescript version to 4.5+ so that we can make use of the latest ecmascript features
* Upgrade `@types/node` to 16.x as well

### Issues Fixed

* Relates #32 - we should bring in the dynamic modules, better tree-shaking, and some compatibility techniques like in https://github.com/MatrixAI/js-id/issues/2

### Tasks

- [x] 1. Update `@types/node` from `14.14.35` to `16.11.7`
- [x] 2. Update `pkg` from `5.3.0` to `5.6.0`
- [x] 3. Update `typescript` from `4.1.3` to `4.5.2`
- [x] 4. Update `pkgs.nix` to https://github.com/NixOS/nixpkgs/commit/a5774e76bb8c3145eac524be62375c937143b80c
- [x] 5. Update tsconfig target from `ES2020` to `ES2021`
- [x] 6. Update `pkgBuilds` from `3.1` to `3.3` with node version `16.14.2`
- [x] 7. Update eslint ecma version from 2020 to 2021
- [x] 8. Update `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` from `4.12.0` to `5.0.0`
- [x] 9. Update `node-gyp-build` from `4.2.3` to `4.4.0`
- ~[ ] 10. Remove leveldown patch for `pkg`, and replace it with `--no-dict leveldown` option as specified here: https://github.com/vercel/pkg/pull/1273~ - this did not work, and also the package.json patch did not work, this is kept the way it is
- [x] 11. Updated typedoc to support TS 4.5.x and regenerated the docs

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased


CMCDragonkai:
Note that the current `pkgs.nix` hash matches what we are using in our platforms. However it currently still hasn't fully updated to node2nix 1.10.0. Initial `nix-shell` currently works, however, we would want to make sure we are using 1.10.0 so we can try out the node 16.x and to do testing. The `pkg` however has been updated to 5.4.1.

CMCDragonkai:
Possible issues:

* https://github.com/NixOS/nixpkgs/pull/146440
* https://github.com/svanderburg/node2nix/issues/274
* https://github.com/NixOS/nixpkgs/issues/145432

CMCDragonkai:
@tegefaulkes relevant for our soon to be release.

CMCDragonkai:
Will need to test master of nixpkgs, not 21.11.

CMCDragonkai:
Node 16 has Promise.any and AggregateError too.

This should be ideally used in our ICE of direct connection and signalling.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

@tegefaulkes @emmacasolin

emmacasolin:
With the following changes I've got `Promise.any` working. All of the existing tests are also still passing.

1. In `package.json` `@types/node` has been updated from `14.14.35` to `16.11.7` - this is the only version of Node 16 supported by `@types/node` https://www.npmjs.com/package/@types/node?activeTab=versions
2. In `package.json` `pkg` has been updated from `5.3.0` to `5.4.1`
3. In `package.json` `typescript` has been updated from `4.1.3` to `4.5.2` - this looks to be the oldest stable version of 4.5 https://github.com/microsoft/TypeScript/tags
4. In `pkgs.nix` `rev` is pointing to `master`
5. In `tsconfig.json` `target` has been updated from `ES2020` to `ES2021`
6. In `utils.nix` `pkgBuilds` has been updated from `3.1` to `3.3` with node version `16.14.2` - this is the only supported node version I could find here https://github.com/vercel/pkg-fetch/tags

CMCDragonkai:
Make sure that the version used by PKG fetch is the same version of nodejs used in nix-shell. Or at the very least a compatible version. The hashes would have to be updated.

CMCDragonkai:
And test that AbortSignal exists and AggregateError exists.

emmacasolin:
> Make sure that the version used by PKG fetch is the same version of nodejs used in nix-shell. Or at the very least a compatible version. The hashes would have to be updated.
>
> And test that AbortSignal exists and AggregateError exists.

They're both using 16.14.2, and I've updated the hashes as well. AbortSignal and AggregateError also both exist, as well as Promise.any.

emmacasolin:
With all of the changes from this comment https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/37#issuecomment-1105875754 both building the releases

```
nix-build ./release.nix --attr application
nix-build ./release.nix --attr docker
```

And installing into Nix user profile

```
nix-env -f ./release.nix --install --attr application
```

Work without issues.

However, I'm getting errors from attempting to build the package

```
[nix-shell:~/Projects/TypeScript-Demo-Lib]$ nix-build -E '(import ./pkgs.nix).callPackage ./default.nix {}'
error: value is a function while a set was expected, at (string):1:1
```

And installing into Docker

```
[nix-shell:~/Projects/TypeScript-Demo-Lib]$ loaded="$(docker load --input "$(nix-build ./release.nix --attr docker)")"

[nix-shell:~/Projects/TypeScript-Demo-Lib]$ image="$(cut -d' ' -f3 <<< "$loaded")"

[nix-shell:~/Projects/TypeScript-Demo-Lib]$ docker run -it "$image"
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/bin/typescript-demo-lib": stat /bin/typescript-demo-lib: no such file or directory: unknown.
```

CMCDragonkai:
After doing this:

```
nix-env -f ./release.nix --install --attr application
```

Are you able to run the executable as `typescript-demo-lib`?

Note that you should try just doing `nix-build ./release.nix --attr application` and then try `./results/bin/typescript-demo-lib` as well. What is in the `result` symlink? Use `tree ./result`.

I read somewhere that node2nix has a problem producing the final executable right now, and that could be causing the docker image problem.

CMCDragonkai:
Btw `nix-build -E '(import ./pkgs.nix).callPackage ./default.nix {}'` is no longer correct. I hadn't updated it yet.

It now needs to be `nix-build -E '(import ./pkgs.nix {}).callPackage ./default.nix {}'`.

CMCDragonkai:
Can you update the README.md with the new command.

CMCDragonkai:
If the bin executable is missing. See if this issue is the culprit: https://github.com/NixOS/nixpkgs/issues/145432.

Furthermore, remember to update the task list for this PR with your changes.

CMCDragonkai:
Remember to push up your changes so I can run it too.

CMCDragonkai:
And if the bin executable is truely missing, then you can try to debug why inside the `default.nix`. See all the `cp` and `ln -s` commands. Those are all the shell commands that run to get the final output correct in `$out/bin`. You can trace backwards by injecting `ls` at various places like we do with `console.log` to see what is happening inside during the build.

```
      # copy the dist
      cp -r ${utils.node2nixDev}/lib/node_modules/${utils.node2nixDev.packageName}/dist $out/lib/node_module>
      # copy over the production dependencies
      if [ -d "${utils.node2nixProd}/lib/node_modules" ]; then
        cp -r ${utils.node2nixProd}/lib/node_modules $out/lib/node_modules/${utils.node2nixDev.packageName}/
      fi
      # create symlink to the deployed executable folder, if applicable
      if [ -d "${utils.node2nixDev}/lib/node_modules/.bin" ]; then
        cp -r ${utils.node2nixDev}/lib/node_modules/.bin $out/lib/node_modules/
        ln -s $out/lib/node_modules/.bin $out/bin # <--------- this is what should be producing the final bin link
      fi
```

emmacasolin:
Running `nix-build ./release.nix --attr application` produces a `result` symlink with just `lib/node_modules/@matrixai/typescript-demo-lib/...`

CMCDragonkai:
Exactly, can you use `tree ./result` to show the full output?

So the bin executable isn't being produced. This is the thing that people have been mentioning for `node2nix`.

Can you debug `default.nix` https://github.com/MatrixAI/TypeScript-Demo-Lib/pull/37#issuecomment-1106057229 to find out why the bin executable isn't being created.

CMCDragonkai:
For reference, this is what the current master shows:

```
»» ~/Projects/TypeScript-Demo-Lib
 ♖ ll result/                                                                        (master) pts/7 17:09:43
total 9.0K
lrwxrwxrwx 2 root root 91 Jan  1  1970 bin -> /nix/store/1z01yfni6z6v6283w203i6pnkiqlw5gq-typescript-demo-lib-1.2.0/lib/node_modules/.bin/
dr-xr-xr-x 3 root root  3 Jan  1  1970 lib/
```

Actually `tree` produces too much data. But the `bin -> ...` is what we are looking for. So you might want to check the pointed-to symlink to see if `lib/node_modules/.bin` exists..

CMCDragonkai:
Then you post an issue upstream on node2nix (the issue I linked was on nixpkgs).

emmacasolin:
Digging into this a bit more, it looks like running any of `nix-build -E '(import ./pkgs.nix {}).callPackage ./default.nix {}'`, `nix-build ./release.nix --attr application`, or `nix-env -f ./release.nix --install --attr application` creates just the `lib` subdirectory

```
[nix-shell:~/Projects/TypeScript-Demo-Lib]$ ll result/
total 9
dr-xr-xr-x 3 root root 3 Jan  1  1970 lib
```

But running either of the docker-related commands completely breaks the symlink:

```
[nix-shell:~/Projects/TypeScript-Demo-Lib]$ tree ./result
./result [error opening dir]

0 directories, 0 files

[nix-shell:~/Projects/TypeScript-Demo-Lib]$ ll result/
ls: cannot access 'result/': Not a directory
```

So I don't think the problem is only with `./default.nix`

CMCDragonkai:
And what is in side `./result/lib/node_modules/.bin`?

emmacasolin:
Nothing, it doesn't exist

CMCDragonkai:
Ok let's see what's inside the result tomorrow. But you may need to go into node2nix generated nix files and see what's going on with the generated package instructions.

emmacasolin:
Looks like the problem might be that pkgs.nix hash I'm using isn't using node2nix 1.10.0 properly. Even though it says it's using 1.10.0, when running `node2nix --version` it says 1.9.0. So I'll have to find a different hash to use.

emmacasolin:
> Looks like the problem might be that pkgs.nix hash I'm using isn't using node2nix 1.10.0 properly. Even though it says it's using 1.10.0, when running `node2nix --version` it says 1.9.0. So I'll have to find a different hash to use.

So unfortunately it looks like even master is still using 1.9.0 https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/node-packages/default.nix#L261

emmacasolin:
Some findings after installing node2nix 1.11.1 (latest) into my own user profile and using it to generate executables:

- Both the version installed into my user profile and the nixpkgs version will generate/overwrite the required files needed to produce executables (`node-packages.nix`, `node-env.nix`, and `default.nix`) and produce an executable with the commands `node2nix` followed by `nix-build -A package`, however I haven't tested this executable on either version
- Both versions also generate the required files and executables with the commands `node2nix -l ./package-lock.json` followed by `nix-build -A package`
- Both versions generate the required files with the commands `node2nix -l ./package-lock.json -16` followed by `nix-build -A package`, however only the nixpkgs version generates an executable (this is probably because -16 isn't a valid option for node2nix 1.9.0, so it just behaves like the option isn't even there)

Although the nixpkgs version generates files and executables using the commands from above, it doesn't generate either (files nor executables) when using the commands from the README

emmacasolin:
It looks like the smallest change that prevents executables from being created is as follows (this should be reproducible using what's currently pushed up on this branch):

1. Run command `node2nix -l ./package-lock.json -16`
2. Open `default.nix` which should have been overwritten - change `<nixpkgs>` on line 3 to `./pkgs.nix`
3. Run command `nix-build -A package`
4. An executable should have been produced in `result/bin` - you can run it but it will fail since it's using the wrong version of node (12_x)
5. Go back to `default.nix` and change the 12 on line 5 to 16
6. Delete the `result` folder and repeat step 3. This time there should not be an executable

CMCDragonkai:
I've changed the `pkgs.nix` to point to nixpkgs upstream rather than our fork. It's actually faster to download (seems to be).

```
import (
  let rev = "a5774e76bb8c3145eac524be62375c937143b80c"; in
  fetchTarball "https://github.com/NixOS/nixpkgs/archive/${rev}.tar.gz"
)
```

CMCDragonkai:
Now to get the master version of node2nix into our nix environment we need to use `fetchFromGitHub`.

To make it easier to know what parameters we need we use `nix-prefetch-git`: `nix-prefetch-git https://github.com/svanderburg/node2nix`.

This tells us:

```
git revision is 68f5735f9a56737e3fedceb182705985e3ab8799
path is /nix/store/4yrqx6jr155r814n2m6vchxy74bda0n4-node2nix
git human-readable version is -- none --
Commit date is 2022-04-20 22:48:58 +0200
hash is 1f791vikig65ly5vcw6zjd0nv2qb8l5w5lr3xy343iq6746s1bil
{
  "url": "https://github.com/svanderburg/node2nix",
  "rev": "68f5735f9a56737e3fedceb182705985e3ab8799",
  "date": "2022-04-20T22:48:58+02:00",
  "path": "/nix/store/4yrqx6jr155r814n2m6vchxy74bda0n4-node2nix",
  "sha256": "1f791vikig65ly5vcw6zjd0nv2qb8l5w5lr3xy343iq6746s1bil",
  "fetchLFS": false,
  "fetchSubmodules": false,
  "deepClone": false,
  "leaveDotGit": false
}
```

Using the `sha256`, we get a `fetchFromGitHub` expression:

```nix
node2nix = fetchFromGitHub { owner = "svanderburg"; repo = "node2nix"; rev = "68f5735f9a56737e3fedceb182705985e3ab8799"; sha256 = "1f791vikig65ly5vcw6zjd0nv2qb8l5w5lr3xy343iq6746s1bil"; };
```

That can be pasted in (minus the `;`) to assign to `node2nix`. Use `:b node2nix` to see.

Now we can refer to files within the repo with `"${node2nix}/release.nix"`.

We know this produces the `package.x86_64-linux` attribute from the `README.md` https://github.com/svanderburg/node2nix#building-a-development-version

So this expression is now possible:

```
(import "${node2nix}/release.nix" {}).package.x86_64-linux
```

Which can be used in `nix repl` and incorporated into our `utils.nix`.

```
nix-repl> :b (import "${node2nix}/release.nix" {}).package.x86_64-linu[2 built, 26 copied (369.0 MiB), 47.2 MiB DL]

this derivation produced the following outputs:
  out -> /nix/store/47gjaiyapflffrvyvxgz57jdy0njn9vw-node2nix-1.11.1
```

Now all of our expressions that use `nodePackages.node2nix` has to be changed to use our custom `node2nix` inside the `utils.nix`.

CMCDragonkai:
Integrating it into our nix files was really easy.

The only major change was for `node2nixDrv` in `utils.nix`.

```nix
  # custom node2nix directly from GitHub
  node2nixSrc = fetchFromGitHub {
    owner = "svanderburg";
    repo = "node2nix";
    rev = "68f5735f9a56737e3fedceb182705985e3ab8799";
    sha256 = "1f791vikig65ly5vcw6zjd0nv2qb8l5w5lr3xy343iq6746s1bil";
  };
  node2nix = (import "${node2nixSrc}/release.nix" {}).package.x86_64-linux;
  node2nixDrv = dev: runCommandNoCC "node2nix" {} ''
    mkdir $out
    ${node2nix}/bin/node2nix \
      ${lib.optionalString dev "--development"} \
      --input ${src}/package.json \
      --lock ${src}/package-lock.json \
      --node-env $out/node-env.nix \
      --output $out/node-packages.nix \
      --composition $out/default.nix \
      --nodejs-${nodeVersion}
  '';
```

It was changed from `nodePackages.node2nix` to just `node2nix`, which can refer to the `node2nix` we just defined due to the `rec { ... }` mark which indicates a mutually recursive attribute set.

CMCDragonkai:
I believe with the new update to `pkg`, we don't need the leveldown patch anymore. See: https://github.com/vercel/pkg/pull/1273.

Furthermore nixpkgs master is now using `pkg` at `5.6.0`. We have to update our `package.json` to match the exact same version too.

CMCDragonkai:
Finally running `nix-build ./release.nix -A application` shows us what has changed:

The build is located at: `/nix/store/mc9714j1lmx15c0781wabsz4rnd1s5ag-typescript-demo-lib-1.2.0`.

```
»» /nix/store/mc9714j1lmx15c0781wabsz4rnd1s5ag-typescript-demo-lib-1.2.0
 ♖ tree -a -L 6 .                                                                               pts/5 17:44:43
.
└── lib
    └── node_modules
        └── @matrixai
            └── typescript-demo-lib
                ├── dist
                │   ├── bin
                │   ├── index.d.ts
                │   ├── index.js
                │   ├── index.js.map
                │   ├── lib
                │   ├── test.json
                │   ├── utils.d.ts
                │   ├── utils.js
                │   └── utils.js.map
                ├── node_modules
                │   ├── abstract-leveldown
                │   ├── base64-js
                │   ├── .bin
                │   ├── buffer
                │   ├── callsites
                │   ├── catering
                │   ├── debug
                │   ├── deferred-leveldown
                │   ├── encoding-down
                │   ├── esm
                │   ├── fd-lock
                │   ├── ieee754
                │   ├── inherits
                │   ├── is-buffer
                │   ├── is-observable
                │   ├── level
                │   ├── level-codec
                │   ├── level-concat-iterator
                │   ├── leveldown
                │   ├── level-errors
                │   ├── level-iterator-stream
                │   ├── level-js
                │   ├── level-packager
                │   ├── level-supports
                │   ├── levelup
                │   ├── ltgt
                │   ├── ms
                │   ├── napi-macros
                │   ├── node-gyp-build
                │   ├── observable-fns
                │   ├── queue-microtask
                │   ├── queue-tick
                │   ├── readable-stream
                │   ├── run-parallel-limit
                │   ├── safe-buffer
                │   ├── string_decoder
                │   ├── threads
                │   ├── timeout-refresh
                │   ├── tiny-worker
                │   ├── unordered-set
                │   ├── util-deprecate
                │   ├── utp-native
                │   └── uuid
                └── package.json

51 directories, 8 files
```

What we can see here is that the `npm build` works fine. Because we do have the `dist/bin` executable of `typescript-demo-lib.js`. In fact it is possible to execute it direclty with the full path:

```
node /nix/store/mc9714j1lmx15c0781wabsz4rnd1s5ag-typescript-demo-lib-1.2.0/lib/node_modules/@matrixai/typescript-demo-lib/dist/bin/typescript-demo-lib.js
```

Or:

```
./result/lib/node_modules/@matrixai/typescript-demo-lib/dist/bin/typescript-demo-lib.js
```

And it works fine.

So the only issue here is why there's no bin symlink in the new node2nix, whether it is caused by `npm` or by `node2nix`.

At the end of the day, we can create our own in `default.nix` to fix.

CMCDragonkai:
The last section of our `default.nix` does this:

```
    # create symlink to the deployed executable folder, if applicable
    if [ -d "${utils.node2nixDev}/lib/node_modules/.bin" ]; then
      cp -r ${utils.node2nixDev}/lib/node_modules/.bin $out/lib/node_modules/
      ln -s $out/lib/node_modules/.bin $out/bin
    fi
```

This is checking if `.bin` directory exists in the `${node2nixDev}/lib/node_modules`. If it is, it copies it into the same relative location, and then creates the final symlink of `$out/bin` to `$out/lib/node_modules/.bin`.

This could mean that `${node2nixDev}/lib/node_modules/.bin` no longer exists.

We should check what is in the `${utils.node2nixDev}` and `${utisl.node2nixProd}`.

CMCDragonkai:
Ok so we can find out what the `node2nixDev` and `node2nixProd` are by just rebuilding and outputting them on the log. By using `-v` flag on our `cp`, `ln` and `mkdir` commands, we can see the paths.

We also see the derivation paths at the beginning of the `nix-build` like:

```
these derivations will be built:
  /nix/store/gfqrdjma9abr8hwv0m73lhxiz741i0d8-_at_matrixai_slash_typescript-demo-lib-1.2.0.drv
  /nix/store/izzrs98ph0l7zgck46xcar1h0j2d4hmy-node-dependencies-_at_matrixai_slash_typescript-demo-lib-1.2.0.drv
  /nix/store/f8f0l5swmwl0jgsxgliafval2w3wisdk-typescript-demo-lib-1.2.0.drv
```

Using `nix show-derivation`, we can see the `output` paths too. And we end up with:

```
# node2nixProd
/nix/store/3kizaya6hmb109lkziin07fr510dir0n-node-dependencies-_at_matrixai_slash_typescript-demo-lib-1.2.0

# node2nixDev
/nix/store/az04yxznn7mf2iigdy03j41i2llr4xbd-_at_matrixai_slash_typescript-demo-lib-1.2.0
```

Looking into these 2 paths, we can see that `node2nixDev` is misisng the `lib/node_modules/.bin` directory.

Now according to my comment in `node2nixDev`, it is the job of `npm install` to actually create symlinks in `lib/node_modules/.bin`.

```
  node2nixDev = (import (node2nixDrv true) { inherit pkgs nodejs; }).package.override (attrs: {
    src = src;
    buildInputs = attrs.buildInputs ++ [ nodePackages.node-gyp-build ];
    dontNpmInstall = true;
    postInstall = ''
      # The dependencies were prepared in the install phase
      # See `node2nix` generated `node-env.nix` for details.
      npm run build

      # This call does not actually install packages. The dependencies
      # are present in `node_modules` already. It creates symlinks in
      # $out/lib/node_modules/.bin according to `bin` section in `package.json`.
      npm install
    '';
  });
```

Perhaps the `npm install` functionality has changed in node16 such that it no longer does this.

CMCDragonkai:
I'm beginning to realise that `npm install` should never have been expected to actually produce that `.bin` directory. At least it only makes sense if `npm install` was installing `@matrixai/typescript-demo-lib` as a dependency in higher-level package.

That is the assumption that I had about how this is supposed to work. Given that the `node2nixDev` produces a `lib/node_modules` that itself contains `@matrixai/typescript-demo-lib`.

But I realised that the `npm install` command would have to be running inside the directory of `typescript-demo-lib`, and if so, it couldn't possibly be creating the `.bin` directory.

On master, this is the output of `node2nixDev`:

```
»» /nix/store/jwp5y8q431nb275q4v7hcpa7b73f55b4-node__at_matrixai_slash_typescript-demo-lib-1.2.0
 ♖ tree -a -L 5 .                                                                                                                                                                          pts/9 18:34:05
.
└── lib
    └── node_modules
        ├── .bin
        │   └── typescript-demo-lib -> ../@matrixai/typescript-demo-lib/dist/bin/typescript-demo-lib.js
        └── @matrixai
            └── typescript-demo-lib
                ├── default.nix
                ├── dist
                ├── docs
                ├── .editorconfig
                ├── .env.example
                ├── .eslintrc
                ├── .github
                ├── .gitignore
                ├── .gitlab-ci.yml
                ├── jest.config.js
                ├── LICENSE
                ├── nix
                ├── node_modules
                ├── .npmignore
                ├── package.json
                ├── package-lock.json
                ├── pkgs.nix
                ├── .prettierrc
                ├── README.md
                ├── release.nix
                ├── shell.nix
                ├── src
                ├── tests
                ├── tsconfig.build.json
                ├── tsconfig.json
                └── utils.nix

12 directories, 20 files
```

It doesn't really make sense that `npm install` that is run within `lib/node_modules/@matrixai/typescript-demo-lib` would create the `.bin` directory located at `lib/node_modules/.bin`.

After all we use `npm install` already in `shell.nix`, and it only installs the bin executables of our dependencies, not our own package.

Perhaps originally `node2nix` did something else that mapped the executable, but because we disabled it's native installation procedure with `dontNpmInstall = true;`, then the new version of `node2nix` also doesn't do the whole creation of that `.bin` directory.

CMCDragonkai:
We can start to play around with the `node2nixDev` directly on `nix repl`.

```
»» ~/Projects/tmp/TypeScript-Demo-Lib
 ♖ nix repl ./pkgs.nix                                                                                                                                                          ⚡(master) pts/9 18:46:05
Welcome to Nix version 2.3.16. Type :? for help.

Loading './pkgs.nix'...
Added 14584 variables.

nix-repl> u = pkgs.callPackage ./utils.nix {}

nix-repl> u.node2nixDev
[1 built, 0.0 MiB DL]«derivation /nix/store/nx97yd5pvckvayagvzf13lha511j3kf0-node__at_matrixai_slash_typescript-demo-lib-1.2.0.drv»

nix-repl> :b u.node2nixDev
```

And we can fiddle with the `node2nixDev` and see how it changes what gets built.

CMCDragonkai:
Unfortunately `nix repl` doesn't show the full logs. Until the `:logs` command is available: https://github.com/NixOS/nix/issues/3504.

So instead we can use `nix-build`:

```
nix-build -E '((import ./pkgs.nix {}).callPackage ./utils.nix {}).node2nixDev'
```

And this will show us the full logs.

CMCDragonkai:
Ok in master, if we disable `npm install`, we get the same result as it is now. An output that doesn't have the `.bin` links.

However in this PR, `npm install` no longer does this.

I think this fairly strange behaviour, in that even back when I discovered this, I thought it was strange that `npm install` would actually produce this `.bin` link.

So I do reckon it's probably a combination of node2nix changing what `npm install` is doing, possibly due to some post installation hook?

I'm going to check the node2nix generated files and compare them.

CMCDragonkai:
Using `nix-build -E '((import ./pkgs.nix {}).callPackage ./utils.nix {}).node2nixDrv true'`.

We get 2 outputs to compare:

```
# this PR
/nix/store/v5l7x96hyzb84jx6lcnzr2qzajby6fv2-node2nix
```

vs

```
# on master
/nix/store/4dlgn3s8l1nvl8q65ff6paalfrkcikz4-node2nix
```

We are looking for post installation hooks that concern themselves with creating symlinks for `.bin`.

CMCDragonkai:
These are the only differences:

```
diff --git a/nix/store/4dlgn3s8l1nvl8q65ff6paalfrkcikz4-node2nix/node-env.nix b/./node-env.nix
index c2b7231..2590dd2 100644
--- a/nix/store/4dlgn3s8l1nvl8q65ff6paalfrkcikz4-node2nix/node-env.nix
+++ b/./node-env.nix
@@ -1,6 +1,6 @@
 # This file originates from node2nix
 
-{lib, stdenv, nodejs, python2, pkgs, libtool, runCommand, writeTextFile}:
+{lib, stdenv, nodejs, python2, pkgs, libtool, runCommand, writeTextFile, writeShellScript}:
 
 let
   # Workaround to cope with utillinux in Nixpkgs 20.09 and util-linux in Nixpkgs master
@@ -40,36 +40,22 @@ let
       '';
     };
 
-  includeDependencies = {dependencies}:
-    lib.optionalString (dependencies != [])
-      (lib.concatMapStrings (dependency:
-        ''
-          # Bundle the dependencies of the package
-          mkdir -p node_modules
-          cd node_modules
-
-          # Only include dependencies if they don't exist. They may also be bundled in the package.
-          if [ ! -e "${dependency.name}" ]
-          then
-              ${composePackage dependency}
-          fi
+  # Common shell logic
+  installPackage = writeShellScript "install-package" ''
+    installPackage() {
+      local packageName=$1 src=$2
 
-          cd ..
-        ''
-      ) dependencies);
+      local strippedName
 
-  # Recursively composes the dependencies of a package
-  composePackage = { name, packageName, src, dependencies ? [], ... }@args:
-    builtins.addErrorContext "while evaluating node package '${packageName}'" ''
-      DIR=$(pwd)
+      local DIR=$PWD
       cd $TMPDIR
 
-      unpackFile ${src}
+      unpackFile $src
 
       # Make the base dir in which the target dependency resides first
-      mkdir -p "$(dirname "$DIR/${packageName}")"
+      mkdir -p "$(dirname "$DIR/$packageName")"
 
-      if [ -f "${src}" ]
+      if [ -f "$src" ]
       then
           # Figure out what directory has been unpacked
           packageDir="$(find . -maxdepth 1 -type d | tail -1)"
@@ -79,28 +65,53 @@ let
           chmod -R u+w "$packageDir"
 
           # Move the extracted tarball into the output folder
-          mv "$packageDir" "$DIR/${packageName}"
-      elif [ -d "${src}" ]
+          mv "$packageDir" "$DIR/$packageName"
+      elif [ -d "$src" ]
       then
           # Get a stripped name (without hash) of the source directory.
           # On old nixpkgs it's already set internally.
           if [ -z "$strippedName" ]
           then
-              strippedName="$(stripHash ${src})"
+              strippedName="$(stripHash $src)"
           fi
 
           # Restore write permissions to make building work
           chmod -R u+w "$strippedName"
 
           # Move the extracted directory into the output folder
-          mv "$strippedName" "$DIR/${packageName}"
+          mv "$strippedName" "$DIR/$packageName"
       fi
 
-      # Unset the stripped name to not confuse the next unpack step
-      unset strippedName
+      # Change to the package directory to install dependencies
+      cd "$DIR/$packageName"
+    }
+  '';
 
-      # Include the dependencies of the package
-      cd "$DIR/${packageName}"
+  # Bundle the dependencies of the package
+  #
+  # Only include dependencies if they don't exist. They may also be bundled in the package.
+  includeDependencies = {dependencies}:
+    lib.optionalString (dependencies != []) (
+      ''
+        mkdir -p node_modules
+        cd node_modules
+      ''
+      + (lib.concatMapStrings (dependency:
+        ''
+          if [ ! -e "${dependency.packageName}" ]; then
+              ${composePackage dependency}
+          fi
+        ''
+      ) dependencies)
+      + ''
+        cd ..
+      ''
+    );
+
+  # Recursively composes the dependencies of a package
+  composePackage = { name, packageName, src, dependencies ? [], ... }@args:
+    builtins.addErrorContext "while evaluating node package '${packageName}'" ''
+      installPackage "${packageName}" "${src}"
       ${includeDependencies { inherit dependencies; }}
       cd ..
       ${lib.optionalString (builtins.substring 0 1 packageName == "@") "cd .."}
@@ -246,8 +257,8 @@ let
           var packageLock = JSON.parse(fs.readFileSync("./package-lock.json"));
 
           if(![1, 2].includes(packageLock.lockfileVersion)) {
-             process.stderr.write("Sorry, I only understand lock file versions 1 and 2!\n");
-             process.exit(1);
+            process.stderr.write("Sorry, I only understand lock file versions 1 and 2!\n");
+            process.exit(1);
           }
 
           if(packageLock.dependencies !== undefined) {
@@ -379,7 +390,7 @@ let
   buildNodePackage =
     { name
     , packageName
-    , version
+    , version ? null
     , dependencies ? []
     , buildInputs ? []
     , production ? true
@@ -391,13 +402,14 @@ let
     , dontStrip ? true
     , unpackPhase ? "true"
     , buildPhase ? "true"
+    , meta ? {}
     , ... }@args:
 
     let
-      extraArgs = removeAttrs args [ "name" "dependencies" "buildInputs" "dontStrip" "dontNpmInstall" "preRebuild" "unpackPhase" "buildPhase" ];
+      extraArgs = removeAttrs args [ "name" "dependencies" "buildInputs" "dontStrip" "dontNpmInstall" "preRebuild" "unpackPhase" "buildPhase" "meta" ];
     in
     stdenv.mkDerivation ({
-      name = "node_${name}-${version}";
+      name = "${name}${if version == null then "" else "-${version}"}";
       buildInputs = [ tarWrapper python nodejs ]
         ++ lib.optional (stdenv.isLinux) utillinux
         ++ lib.optional (stdenv.isDarwin) libtool
@@ -414,6 +426,8 @@ let
       passAsFile = [ "compositionScript" "pinpointDependenciesScript" ];
 
       installPhase = ''
+        source ${installPackage}
+
         # Create and enter a root node_modules/ folder
         mkdir -p $out/lib/node_modules
         cd $out/lib/node_modules
@@ -427,6 +441,14 @@ let
         if [ -d "$out/lib/node_modules/.bin" ]
         then
             ln -s $out/lib/node_modules/.bin $out/bin
+
+            # Patch the shebang lines of all the executables
+            ls $out/bin/* | while read i
+            do
+                file="$(readlink -f "$i")"
+                chmod u+rwx "$file"
+                patchShebangs "$file"
+            done
         fi
 
         # Create symlinks to the deployed manual page folders, if applicable
@@ -446,13 +468,18 @@ let
         # Run post install hook, if provided
         runHook postInstall
       '';
+
+      meta = {
+        # default to Node.js' platforms
+        platforms = nodejs.meta.platforms;
+      } // meta;
     } // extraArgs);
 
   # Builds a node environment (a node_modules folder and a set of binaries)
   buildNodeDependencies =
     { name
     , packageName
-    , version
+    , version ? null
     , src
     , dependencies ? []
     , buildInputs ? []
@@ -470,7 +497,7 @@ let
       extraArgs = removeAttrs args [ "name" "dependencies" "buildInputs" ];
     in
       stdenv.mkDerivation ({
-        name = "node-dependencies-${name}-${version}";
+        name = "node-dependencies-${name}${if version == null then "" else "-${version}"}";
 
         buildInputs = [ tarWrapper python nodejs ]
           ++ lib.optional (stdenv.isLinux) utillinux
@@ -486,6 +513,8 @@ let
         passAsFile = [ "includeScript" "pinpointDependenciesScript" ];
 
         installPhase = ''
+          source ${installPackage}
+
           mkdir -p $out/${packageName}
           cd $out/${packageName}
 
@@ -498,6 +527,7 @@ let
             if [ -f ${src}/package-lock.json ]
             then
                 cp ${src}/package-lock.json .
+                chmod 644 package-lock.json
             fi
           ''}
 
@@ -520,7 +550,7 @@ let
   buildNodeShell =
     { name
     , packageName
-    , version
+    , version ? null
     , src
     , dependencies ? []
     , buildInputs ? []
@@ -536,9 +566,10 @@ let
 
     let
       nodeDependencies = buildNodeDependencies args;
+      extraArgs = removeAttrs args [ "name" "dependencies" "buildInputs" "dontStrip" "dontNpmInstall" "unpackPhase" "buildPhase" ];
     in
-    stdenv.mkDerivation {
-      name = "node-shell-${name}-${version}";
+    stdenv.mkDerivation ({
+      name = "node-shell-${name}${if version == null then "" else "-${version}"}";
 
       buildInputs = [ python nodejs ] ++ lib.optional (stdenv.isLinux) utillinux ++ buildInputs;
       buildCommand = ''
@@ -557,7 +588,7 @@ let
         export NODE_PATH=${nodeDependencies}/lib/node_modules
         export PATH="${nodeDependencies}/bin:$PATH"
       '';
-    };
+    } // extraArgs);
 in
 {
   buildNodeSourceDist = lib.makeOverridable buildNodeSourceDist;
```

I can't find any usage mention of creating the `.bin` directory, so it must be the way `npm install` command is being used.

At any case it seems expecting `npm install` to do this is quite flaky, and perhaps it's better idea to just explicitly acquire the bin of the repo in a different way. Maybe by running `npm install` in the parent directory?

CMCDragonkai:
I've compared the old 1.9.0 to 1.11.0. I suspect that something changed with `npm install` behaviour coupled with changes to `node2nix` `node-env.nix` as per above has led to the `.bin` directory no longer being created.

While upstream figures out why. The workaround right now is to directly refer to these executables under our package and just create those symlinks in `default.nix`. We do this already in our `default.nix`, we just need to change:

```
    # create symlink to the deployed executable folder, if applicable
    if [ -d "${utils.node2nixDev}/lib/node_modules/.bin" ]; then
      cp -v -r ${utils.node2nixDev}/lib/node_modules/.bin $out/lib/node_modules/
      ln -v -s $out/lib/node_modules/.bin $out/bin
    fi
```

To instead create symlinks for every file in: `$out/lib/node_modules/${utils.node2nixDev.packageName}/dist/bin` to `$out/bin`.

Or extract this information from `package.json` bin.

Or continue figuring out how to make `npm install` do what it used to do based on this information: https://docs.npmjs.com/cli/v8/configuring-npm/folders#more-information

CMCDragonkai:
The closest thing I can get to is replacing `npm install` in `node2nixDev` with:

```
npm_config_prefix="$out" npm install --global
```

However this doesn't work, it has a weird error about not finding the `package.json`.

Instead `npm config_prefix="$out/lib" npm install --global` seems to work.

However this doesn't produce the links the place we want to have it. Partly because `npm install` doesn't just setup the `.bin` but also wants to setup symlinks for `node_modules`, and because the name is the same, it results in some problems.

I think our `node2nixDev` might be need to be slightly changed. It may need to use `--install-links` as part of `npm install`. Or we need to override the entire `installPhase`.

CMCDragonkai:
Given that there's no elegant solution atm, I've decided to go with some duct tape until upstream fixes the problem.

In the `default.nix`, at the very end it is now using `jq` to parse out the `package.json` `bin` properties, and automatically generates the symlinks to the bin paths, while also making those bin paths executable with `chmod a+x`.

This works well enough for now, as the previous system was already kind of weird, relying on `npm install` to build symlinks and then mapping them with further symlinks.

Future solutions may involve redoing the node2nix derivations to be more direct to our infrastructure.

CMCDragonkai:
@emmacasolin 5.6.0 of `pkg` is still trying to load v14 nodejs base binaries.

```
> pkg@5.6.0
> Fetching base Node.js binaries to PKG_CACHE_PATH
  fetched-v14.19.1-linux-x64          [] 0%> Not found in remote cache:
  {"tag":"v3.3","name":"node-v14.19.1-linux-x64"}
```

I wonder why this is the case.

Btw if you want ot make sure to get the right hash, use `nix-prefetch-url` on those URLs. Don't just use the ones provided.

CMCDragonkai:
Everything seems to be working now, EXCEPT the docker builds... will wait and see.

CMCDragonkai:
I've tested the linux executable on Ubuntu so it works.

Going to merge this as CI/CD will be able to test on docker and macos builds and windows builds.

The docker problem might be just my computer. Let's see.

tegefaulkes:
I can't seem to build or nix-shell this. I get the error
```
> nix-shell
error: AWS error fetching 'ymcglmwrkg8kdxpp0wnd29y2m17js9ah.narinfo': The AWS Access Key Id you provided does not exist in our records.
```

SO i'll have to check my AWS access. But I don't have this problem with the js-polykey repo.

CMCDragonkai:
It could be that our NixOS has changed something with respect to aws access. That's also about a key id, not the AWS authentication key.

Unfortunate since you'll need to update your NixOS to be the same as all of us, but there's a problem with the current NixOS version we are on and running nodejs inside a docker container.

tegefaulkes:
Trying to run the docker images give me
```
> sudo docker run -it typescript-demo-lib-1.3.0:x02my25yw5hip2cpcrs37iadzhqx8p8c                                                                                              
/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node[1]: ../src/node_platform.cc:61:std::unique_ptr<long unsigned int> node::WorkerThreadsTaskRunner::DelayedTaskScheduler::Start(): Assertion `(0) == (uv_thread_create(t.get(), start_thread, this))' failed.
 1: 0xa64dd8 node::Abort() [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 2: 0xa64e67  [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 3: 0xae5f75 node::WorkerThreadsTaskRunner::WorkerThreadsTaskRunner(int) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 4: 0xae60a2 node::NodePlatform::NodePlatform(int, v8::TracingController*) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 5: 0xa24d2f node::V8Platform::Initialize(int) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 6: 0xa2304b node::InitializeOncePerProcess(int, char**, node::InitializationSettingsFlags, node::ProcessFlags::Flags) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 7: 0xa231d9 node::InitializeOncePerProcess(int, char**) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 8: 0xa23238 node::Start(int, char**) [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]
 9: 0x7f4c495421d7  [/nix/store/ayrsyv7npr0lcbann4k9lxr19x813f0z-glibc-2.34-115/lib/libc.so.6]
10: 0x7f4c49542297 __libc_start_main [/nix/store/ayrsyv7npr0lcbann4k9lxr19x813f0z-glibc-2.34-115/lib/libc.so.6]
11: 0x989ce1 _start [/nix/store/zl4bvsqfxyx5vn9bbhnrmbmpfvzqj4gd-nodejs-16.14.2/bin/node]

```

CMCDragonkai:
I've tested the same image on Ubuntu VM. It works on Ubuntu no problem, and on CI/CD.

So it looks like this is a NixOS problem.

image

CMCDragonkai:
For now since it works in CI/CD and in ubuntu, we should conclude that our docker image is correctly built. But NixOS docker runtime has some misconfiguration/bug. So we will continue as is, until upstream figures out what's going on in our NixOS docker runtime.

CMCDragonkai:
I actually forgot about https://github.com/NixOS/nixpkgs/issues/170279 that we encountered before, so for now the docker builds still fail in NixOS. So I'm going to assume that the docker builds do work.

PR #71, opened 2022-07-10
Feature 22.05

CMCDragonkai:
### Description

Upgrades to 22.05 revision ce6aa13369b667ac2542593170993504932eb836 and adds the `created = "now";`.

### Issues Fixed

* Related https://github.com/MatrixAI/js-polykey/pull/396

### Tasks

- [x] 1. Changed `pkgs.nix` to use the 22.05 revision as are using on our OS
- [x] 2. Added `created = "now";` that introduces proper timestamps on our docker images, but binary reproducibility (content hashes) will change, input hashing is preserved 

### Final checklist

* [x] Domain specific tests
* [x] Full tests
* [x] Lint fixed
* [x] Squash and rebased

### DIFF
--- a/pkgs.nix
+++ b/pkgs.nix
@@ -1,4 +1,4 @@
 import (
-  let rev = "a5774e76bb8c3145eac524be62375c937143b80c"; in
+  let rev = "ce6aa13369b667ac2542593170993504932eb836"; in
   builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/${rev}.tar.gz"
 )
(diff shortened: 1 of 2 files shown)


CMCDragonkai:
The new gitlab-runner and matching revisions give us this result:

We are now only downloading 68 MiB vs originally 239 MiB on the TS demo lib `check:lint` job.

image

image

* The job logs are much shorter
* This particular linux job is pretty small, so overall clock time didn't change much, instead... an increased time was spent pulling the container image from the registry

However less bandwidth used, and I'm sure this will scale up to larger projects in the future, plus we have a more comprehensive image with all the relevant configuration with timezones, TLS, iana, locality... and more.

